PR #2, opened 2023-04-19
Set evaluate_before_training at beginning if epochs==0

### LINKED ISSUE #1: [BUG] Progress is temporarily above 100%
### 🐛 Bug
Progress status gets temporarily above 100%.

Screenshot from 2023-04-18 17-06-33

### To Reproduce

Attached, you find the config used for training. It is using 0 epochs which may cause the issue
[logs_umber-chipmunk.zip](https://github.com/h2oai/h2o-llmstudio/files/11263370/logs_umber-chipmunk.zip)

maxjeblick:
If `epochs ==0` and `Evaluate Before Training` is disabled, evaluation epochs are calculated incorrectly (see [here](https://github.com/h2oai/h2o-llmstudio/blob/main/train.py#L525)).

This PR sets  `cfg.training.evaluate_before_training=True` at the beginning of an experiment.
This PR fixes #1.

### DIFF
--- a/train.py
+++ b/train.py
@@ -432,6 +432,11 @@ def run(cfg: Any) -> None:
     if cfg.training.epochs >= 0:
         do_run_train = True
 
+    # Force evaluation if user trains 0 epochs
+    cfg.training.evaluate_before_training = (
+        cfg.training.evaluate_before_training or cfg.training.epochs == 0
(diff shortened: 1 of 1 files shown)


psinger (approved):
LGTM!

PR #14, opened 2023-04-20
Add default values and stop tokens

psinger:
This PR adds a few default stop tokens used in the field and adds stablelm to the list of default backbones.

### DIFF
--- a/llm_studio/python_configs/text_causal_language_modeling_config.py
+++ b/llm_studio/python_configs/text_causal_language_modeling_config.py
@@ -313,12 +313,12 @@ class ConfigNLPCausalLMTokenizer(DefaultConfig):
 class ConfigNLPCausalLMArchitecture(DefaultConfig):
     model_class: Any = text_causal_language_modeling_model.Model
     pretrained: bool = True
-    pretrained_weights: str = ""
 
     backbone_dtype: str = "float16"
     gradient_checkpointing: bool = False
     force_embedding_gradients: bool = False
     intermediate_dropout: float = 0
+    pretrained_weights: str = ""
 
     def __post_init__(self):
         super().__post_init__()
@@ -335,7 +335,6 @@ class ConfigNLPCausalLMArchitecture(DefaultConfig):
 
         self._visibility["model_class"] = -1
         self._visibility["pretrained"] = -1
-        self._visibility["pretrained_weights"] = -1
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


maxjeblick:
Thanks for this PR!

I have a two questions regarding the stopping criteria:
- [StoppingCriteriaSub](https://github.com/h2oai/h2o-llmstudio/blob/main/llm_studio/src/models/text_causal_language_modeling_model.py#L21) seem to trigger stopping on the first item of the batch. If so, does it make sense to add the stopping criteria for inference batch size == 1, only?
- In the if-else clause [here](https://github.com/h2oai/h2o-llmstudio/blob/6bf496141f9ab8088b2bfda652598e21c6e7afe5/llm_studio/src/datasets/text_utils.py#L66), you probably want to always set `add_special_tokens=False`, as you want to stop on the stop token, alone. In practise, this probably doesn't make a difference, as causal language models usually do not prepend the sequence with a cls token.

maxjeblick (changes requested):
Thanks for this PR!

I have a two questions regarding the stopping criteria:
- [StoppingCriteriaSub](https://github.com/h2oai/h2o-llmstudio/blob/main/llm_studio/src/models/text_causal_language_modeling_model.py#L21) seem to trigger stopping on the first item of the batch. If so, does it make sense to add the stopping criteria for inference batch size == 1, only?
- In the if-else clause [here](https://github.com/h2oai/h2o-llmstudio/blob/6bf496141f9ab8088b2bfda652598e21c6e7afe5/llm_studio/src/datasets/text_utils.py#L66), you probably want to always set `add_special_tokens=False`, as you want to stop on the stop token, alone. In practise, this probably doesn't make a difference, as causal language models usually do not prepend the sequence with a cls token.

maxjeblick:
> Not sure I follow here. I was considering this if/else in case the stop token is part of the special tokens list, and then add_special_tokens=False would not return the token. Or is this not the case?

Tokenization with `add_special_tokens=False` will tokenize and convert to `input_ids`, it just doesn't add additional special tokens. As an example, for `bert-base` models, `add_special_tokens=True` will tokenize the input text and then add token_ids for `[CLS]` to the left and `[SEP]` token to the right.


> How in general is the generate function stopping with EOS tokens with larger batches?

The stopping criteria could be modified, e.g. stop generation if all sequences contain at least one stop token.

psinger:
> The stopping criteria could be modified, e.g. stop generation if all sequences contain at least one stop token.

Yes right, but it will be tricky to post-process, because need to remove everything after the stop tokens.
Maybe batch_size=1 is the most sensible solution for now...

maxjeblick (approved):
LGMT thanks

PR #32, opened 2023-04-23
Tokenizer rework

psinger:


maxjeblick (changes requested):
Thanks for the PR, added functionality looks good.

I'd remove `StoppingCriteriaSub`.  Currently, `text_prompt_start` will set inference batch size to 1 which may be unexpected and even be slower than batched inference without stopping criteria. A stopping criteria working with batches could be added in a subsequent PR.

psinger:
> Thanks for the PR, added functionality looks good.
> 
> I'd remove `StoppingCriteriaSub`. Currently, `text_prompt_start` will set inference batch size to 1 which may be unexpected and even be slower than batched inference without stopping criteria. A stopping criteria working with batches could be added in a subsequent PR.

But stopping criteria is needed. Otherwise the outputs might not be as expected.

maxjeblick:
> But stopping criteria is needed. Otherwise the outputs might not be as expected.

If it is not applied, I'd assume the output is right padded with `[EOS]` token, which would then be discarded while decoding ([here](https://github.com/h2oai/h2o-llmstudio/blob/main/llm_studio/src/datasets/text_causal_language_modeling_ds.py#L157))?

maxjeblick:
I'm also a bit concerned with `get_tokenizer` being able to change inference batch size ([here](https://github.com/h2oai/h2o-llmstudio/blob/main/llm_studio/src/datasets/text_utils.py#L77), not changed in your PR). This has the chance to introduce bugs at some point, e.g. [total_validation_steps](https://github.com/h2oai/h2o-llmstudio/blob/main/train.py#L519) may become incorrect if order in `train.py` is changed at some point.

psinger:
> > But stopping criteria is needed. Otherwise the outputs might not be as expected.
> 
> If it is not applied, I'd assume the output is right padded with `[EOS]` token, which would then be discarded while decoding ([here](https://github.com/h2oai/h2o-llmstudio/blob/main/llm_studio/src/datasets/text_causal_language_modeling_ds.py#L157))?

Yeah, but the generate function has the tendency to not necessarily stop at prompting tokens. So it might generate multiple chained conversations if not stopped at the appropriate tokens.

psinger:
> I'm also a bit concerned with `get_tokenizer` being able to change inference batch size ([here](https://github.com/h2oai/h2o-llmstudio/blob/main/llm_studio/src/datasets/text_utils.py#L77), not changed in your PR). This has the chance to introduce bugs at some point, e.g. [total_validation_steps](https://github.com/h2oai/h2o-llmstudio/blob/main/train.py#L519) may become incorrect if order in `train.py` is changed at some point.

It is changed before subsequent code, not sure what you are concerned about exactly. Could you elaborate please?

maxjeblick:
> Yeah, but the generate function has the tendency to not necessarily stop at prompting tokens. So it might generate multiple chained conversations if not stopped at the appropriate tokens.

Ok, I get your point. We can also merge this PR and rework the stopping criteria to work with batches (shouldn't be too hard, I hope).

psinger:
> > Yeah, but the generate function has the tendency to not necessarily stop at prompting tokens. So it might generate multiple chained conversations if not stopped at the appropriate tokens.
> 
> Ok, I get your point. We can also merge this PR and rework the stopping criteria to work with batches (shouldn't be too hard, I hope).

I am not too convinced it is easy...
We could at least change default inference batch size to 1 then...to make it more transparent

maxjeblick:
```

import logging

import torch
from transformers import StoppingCriteria

logger = logging.getLogger(__name__)


class StopWordCriteria(StoppingCriteria):
    def __init__(self, cfg, stop_word_ids=None):
        super().__init__()
        if stop_word_ids is None:
            stop_word_ids = []

        self.stop_word_ids = [stop_word_id.to(cfg.environment._device) for stop_word_id in stop_word_ids]
        if cfg.environment._local_rank == 0:
            logger.info(f"Stopping criteria tokens: {self.stop_word_ids}")

    def should_stop(self, input_ids: torch.LongTensor, stop_word_id: torch.FloatTensor):
        return (torch.mean(((input_ids == stop_word_id).sum(1) > 0).float()) == 1).item()

    def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor):
        for stop_word_id in self.stop_word_ids:
            if self.should_stop(input_ids, stop_word_id):
                return True

        return False

```

Something like this?

maxjeblick:
```
import logging

import torch
from transformers import StoppingCriteria

logger = logging.getLogger(__name__)


class StoppingCriteriaSub(StoppingCriteria):
    def __init__(self, cfg, stop_word_ids=None):
        super().__init__()
        if stop_word_ids is None:
            stop_word_ids = []

        self.stop_word_ids = [stop_word_id.to(cfg.environment._device) for stop_word_id in stop_word_ids]
        if cfg.environment._local_rank == 0:
            logger.info(f"Stopping criteria tokens: {self.stop_word_ids}")

    def should_stop(self, input_ids: torch.LongTensor, stop_word_id: torch.FloatTensor):
        return (torch.mean(((input_ids == stop_word_id).sum(1) > 0).float()) == 1).item()

    def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor):
        for stop_word_id in self.stop_word_ids:
            if self.should_stop(input_ids, stop_word_id):
                return True

        return False

```

Something like this?

It still requires to postprocess the predicted text in decoding to avoid keeping generated tokens that come what any stop token.

psinger:
The issue with this is that we also have the prompt in the input, so the tokens can exist in the input >= 1 times.

maxjeblick:
> The issue with this is that we also have the prompt in the input, so the tokens can exist in the input >= 1 times.

Yes, that's correct, I missed that. One needs to rework the code, i.e. left trim `input_ids` by `len(batch["prompt_input_ids"])` within the call function and pass `len(batch["prompt_input_ids"])` to the stopping criteria init.

maxjeblick (approved):
LGTM,, thanks! We can merge this PR and rework token generation in a subsequent, decoupled PR.

Issue #17, opened 2023-04-20
[FEATURE] Support nested tree conversation data


psinger:
### 🚀 Feature

Support tree-like conversation data - i.e. chain of thoughts such as the OASST data provdes.

### Motivation

Currently, we only support prompt/output data structures. While one can manually add previous conversations to the input dataframe, it would be very helpful to support this out of the box.

This will be specifically helpful for conversational bots with history.

### Proposed features & solution

A first part of the solution is support providing such information in the input data. I am proposing the following:

1. Allow to set a `parent_column` in the dataset. This column will allow to link individual conversations with each other.
2. Have potentially two extra settings - could be also only one:
    - Probability to link conversations together while training
    - Number of conversations to link together
 
Additionally, I would suggest to add an augmentation setting:
- Random probability to link any random conversations together.

This might help to differentiate between unrelated and related context.

Potential additional tasks:
- [ ] Update README
- [ ] Update configs
- [ ] Update Kaggle Dataset notebook

PR #40, opened 2023-04-24
Chained conversation support

psinger:
This PR addresses a few issues.

First, and foremost, it closes https://github.com/h2oai/h2o-llmstudio/issues/17
It is now possible to add a `parent_column` which maps to a separate `id` column in the data and then chains the samples.

There are also two additional augmentation settings:
`skip_parent_probability` randomly skips chained samples in training
`random_parent_probability` randomly prepends any random sample to a record

For inference, full hierarchy is prepended, or nothing if not available

New data preparation was added to support this functionality.

Additionally, this PR fixes the following issues identified while implementing this PR:
- Neptune logging did not work while re-saving the cfg
- `force_embedding_gradients` was buggy and should be fixed now

It adds also another new `add_prompt_answer_tokens` setting to manually specify whether to add separator tokens to the tokenizer.

### DIFF
--- a/llm_studio/python_configs/text_causal_language_modeling_config.py
+++ b/llm_studio/python_configs/text_causal_language_modeling_config.py
@@ -239,6 +239,8 @@ class ConfigNLPCausalLMDataset(DefaultConfig):
 
     prompt_column: Tuple[str, ...] = ("instruction", "input")
     answer_column: str = "output"
+    parent_column: str = "None"
+
     text_prompt_start: str = ""
     text_answer_separator: str = "\\n"
 
@@ -268,10 +270,13 @@ class ConfigNLPCausalLMDataset(DefaultConfig):
         self._possible_values["data_sample"] = (0.05, 1, 0.05)
         self._possible_values["data_sample_choice"] = ["Train", "Validation"]
         self._possible_values["prompt_column"] = possible_values.Columns(
-            prefer_with=lambda column: column in ("Instruction", "prompt")
+            prefer_with=lambda column: column in ("instruction", "prompt")
         )
         self._possible_values["answer_column"] = possible_values.Columns(
             prefer_with=lambda column: column in ("answer", "output")
+        )
+        self._possible_values["parent_column"] = possible_values.Columns(
+            prefer_with=lambda column: column in ("parent"), add_none=True
         )
 
         self._nesting.add(
@@ -342,10 +347,12 @@ class ConfigNLPCausalLMArchitecture(DefaultConfig):
 class ConfigNLPAugmentation(DefaultConfig):
     nlp_augmentations_class: Any = BaseNLPAug
     token_mask_probability: float = 0
+    random_parent_sample_probability: float = 0
 
     def __post_init__(self):
         super().__post_init__()
         self._possible_values["token_mask_probability"] = (0, 0.9, 0.05)
+        self._possible_values["random_parent_sample_probability"] = (0, 1.0, 0.05)
         self._visibility["nlp_augmentations_class"] = -1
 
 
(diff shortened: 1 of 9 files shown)


llm_studio/src/datasets/text_causal_language_modeling_ds.py
```diff
@@ -65,6 +65,17 @@ def __init__(self, df: pd.DataFrame, cfg: Any, mode: str = "train"):
...
+        self.parent_ids = None
+        if self.cfg.dataset.parent_column != "None":
+            if "id" not in self.df.columns:
+                logger.warning(
+                    f"When using parent column, the dataframe requires an 'id' column. "
+                    f"Disabling functionality for mode {self.mode}."
+                )
+            else:
+                self.parent_ids = self.df[self.cfg.dataset.parent_column].values
+                self.id_to_idx = {v: k for k, v in enumerate(self.df["id"].values)}
```
maxjeblick, line 77: Variable name is a bit generic, maybe `self.df_id_to_idx`?

train.py
```diff
@@ -544,9 +545,11 @@ def run(cfg: Any) -> None:
...
 
     if cfg.architecture.force_embedding_gradients:
-        for param in model.backbone.base_model.get_input_embeddings().parameters():
-            param = param.float()
-            param.requires_grad = True
+        for module in model.modules():
+            if isinstance(module, torch.nn.Embedding):
+                for param in module.parameters():
+                    param.requires_grad = True
+                    param.data = param.data.float()
```
maxjeblick, line 552: What's the reason to cast to float explicitly?

psinger: Mixed precision can't operate on float16 weights, so it will fail. I could only implement it for LORA+MP, but I think for stability it does not hurt to always cast this to float in this case.

llm_studio/src/datasets/text_causal_language_modeling_ds.py
```diff
@@ -65,6 +65,17 @@ def __init__(self, df: pd.DataFrame, cfg: Any, mode: str = "train"):
             self.df[self.cfg.dataset.answer_column].astype(str).values.tolist()
         )
 
+        self.parent_ids = None
+        if self.cfg.dataset.parent_column != "None":
```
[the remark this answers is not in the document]
psinger: Probably it will yes

llm_studio/python_configs/text_causal_language_modeling_config.py
```diff
@@ -20,6 +20,78 @@
...
+        self._possible_values["data_sample"] = (0.05, 1, 0.05)
+        self._possible_values["data_sample_choice"] = ["Train", "Validation"]
+        self._possible_values["prompt_column"] = possible_values.Columns(
+            prefer_with=lambda column: column in ("instruction", "prompt")
+        )
+        self._possible_values["answer_column"] = possible_values.Columns(
+            prefer_with=lambda column: column in ("answer", "output")
+        )
+        self._possible_values["parent_column"] = possible_values.Columns(
+            prefer_with=lambda column: column in ("parent"), add_none=True
```
maxjeblick, line 75: `("parent", )`

llm_studio/python_configs/text_causal_language_modeling_config.py
```diff
@@ -20,6 +20,78 @@
...
+    validation_strategy: str = "automatic"
+    validation_dataframe: str = ""
+    validation_size: float = 0.01
+
+    data_sample: float = 1.0
+    data_sample_choice: Tuple[str, ...] = ("Train", "Validation")
+
+    prompt_column: Tuple[str, ...] = ("instruction", "input")
+    answer_column: str = "output"
+    parent_column: str = "None"
```
maxjeblick, line 38: Maybe rename this to `parent_id_column` ? The default dataset also uses `parent_id` as a column name

maxjeblick (approved):
LGTM, thanks for implementing this nice feature!
I've left some suggestions to rename variables.

PR #42, opened 2023-04-25
Batch implementation for stop tokens

maxjeblick:
This PR adds a batch implementation for stop token usage in `.generate`
- `.generate` will stop if one stop token occurs at least once in every sample of the batch
- `clean_output` will remove any text that comes after a stop token (together with the stop token itself).

### DIFF
--- a/llm_studio/src/datasets/text_causal_language_modeling_ds.py
+++ b/llm_studio/src/datasets/text_causal_language_modeling_ds.py
@@ -175,12 +175,13 @@ class CustomDataset(Dataset):
     ):
         output["predicted_text"] = output["predicted_text"].tolist()
         for j in range(len(output["predicted_text"])):
-            curr_text = output["predicted_text"][j].strip()
+            # Add " " at beginning and end to not miss stop tokens at beginning and end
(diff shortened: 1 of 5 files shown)


psinger:
@maxjeblick 

Started testing it and got this prediction:

image

To replicate:

image

Backbone: togethercomputer/GPT-NeoXT-Chat-Base-20B

psinger:
A bit hard to replicate...as it does not happen always

maxjeblick:
Thanks for the catch,
I tried to fix the issue by checking also for `\nstop_token` in the text.
The easiest way code wise would be to check for stop token directly (`text.find(stop_token) >0`). May be a good solution, too, as stop tokens are not expected to be included in regular words.

psinger:
I think I would just remove it not caring if it is part of a regular word, wdyt?

maxjeblick:
> I think I would just remove it not caring if it is part of a regular word, wdyt?

Yeah, it's the easiest one, I've changed it accordingly.

psinger (approved):
Thanks a lot @maxjeblick - I believe it looks good now!

Issue #43, opened 2023-04-25
[BUG] Neptune logging not working


psinger:
### 🐛 Bug

It appears the new UX upgrade in Neptune has broken the logging, or it is another issue.

Logs on Neptune are empty.

Suggestion is to upgrade the client and try using `append` instead of `log` as described in the docs.

PR #48, opened 2023-04-25
Fix Neptune

psinger:
Closes https://github.com/h2oai/h2o-llmstudio/issues/43

Updates also Neptune package and code


pascal-pfeiffer (approved):
lgtm!

could make sense to print the Exception of the logger [here](https://github.com/h2oai/h2o-llmstudio/blob/93754c04b44ad791c7728663d6710bc7ed1527e8/llm_studio/src/loggers.py#L145):
```
        try:
            self.loggers["external"] = self.loggers["external"](cfg)
        except Exception:
            logger.warning(
                "Error when initializing logger. "
                "Disabling custom logging functionality. "
                "Please ensure logger configuration is correct and "
                "you have a stable Internet connection."
            )
            self.loggers["external"] = DummyLogger(cfg)
```

If someone updates the same env, both, `neptune` and `neptune-client` will be installed and raise an error.

PR #49, opened 2023-04-25
introducing busy_dialog for waiting operations

fatihozturkh2o:
This PR
- Introduces `busy_dialog` under `wave_utils.py`. Instead of defining `q.page["meta"].dialog `+ `q.page.save()` handling everywhere, one can use this as a one-liner and just pass title&text info
- Calling `busy_dialog` when submitting models to HF hub. Seeing the "loading" wheel is not as informative and it's easy to assume the app might have gotten stuck since it takes quite a bit of time to upload LLM models. 
- ~~Calling `busy_dialog` when waiting for the chatbot answers as well. Especially for the chatbots that are slow and the prompts requiring long answers, seeing the dialog can be more informative than seeing the "loading" wheel. (Best would be seeing a loading chat bubble though. Wave doesn't support it yet I guess?)~~
- The usage of busy_dialog can be extended in the code base to have a consistent blocker dialog behavior across the pages/operations
 
Wonder what you think

### How it looks:

https://user-images.githubusercontent.com/70659545/234353838-a773ebb4-5e6d-4a39-be5d-55363a4138e7.mov

vs

Screenshot 2023-04-25 at 20 03 54

### DIFF
--- a/app_utils/wave_utils.py
+++ b/app_utils/wave_utils.py
@@ -359,3 +359,19 @@ async def report_error(q: Q):
     q.page[card_name].items[13].text_xs.visible = True
 
     await q.page.save()
+
+
+async def busy_dialog(q: Q, title: str = "", text: str = "") -> None:
+    """Creates busy dialog"""
+
+    q.page["meta"].dialog = ui.dialog(
+        title=title,
+        primary=True,
+        items=[
+            ui.progress(label=text),
+        ],
+        blocking=True,
+    )
+    await q.page.save()
+    await q.sleep(0.1)
+    q.page["meta"].dialog = None
--- a/app_utils/wave_utils.py
+++ b/app_utils/wave_utils.py
@@ -361,7 +361,9 @@ async def report_error(q: Q):
     await q.page.save()
 
 
-async def busy_dialog(q: Q, title: str = "", text: str = "") -> None:
+async def busy_dialog(
+    q: Q, title: str = "", text: str = "", force_wait: bool = False
+) -> None:
     """Creates busy dialog"""
 
     q.page["meta"].dialog = ui.dialog(
@@ -373,5 +375,6 @@ async def busy_dialog(q: Q, title: str = "", text: str = "") -> None:
         blocking=True,
     )
     await q.page.save()
-    await q.sleep(0.1)
+    if force_wait:
+        await q.sleep(1)
     q.page["meta"].dialog = None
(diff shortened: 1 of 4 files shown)


app_utils/wave_utils.py
```diff
@@ -359,3 +359,19 @@ async def report_error(q: Q):
...
+    q.page["meta"].dialog = ui.dialog(
+        title=title,
+        primary=True,
+        items=[
+            ui.progress(label=text),
+        ],
+        blocking=True,
+    )
+    await q.page.save()
+    await q.sleep(0.1)
```
pascal-pfeiffer, line 376: what is the sleep for?

pascal-pfeiffer (commented):
Thank you @fatihozturkh2o !
I like it a lot for the push to huggingface. For the chat, I am not really sure if it makes much sense, as I find it a bit annoying when it only flashes up for a fraction of a second. 
Can we add three dots maybe to the chat? Or a "generating answer ..." that is being replaced by the actual answer when done?

app_utils/wave_utils.py
psinger: Replacing messages does not work in the wave chat element unfortunately.

fatihozturkh2o: @pascal-pfeiffer thanks for the feedback!

About the chat, I agree that having a "..." bubble within the chat would be the best while waiting for the answer. However, to my knowledge, Wave does not support this yet and more importantly, since operations like "model.generate" are running on the main process, the app serve function gets stuck anyway. 

And for the current example, I agree it looks like chat dialog pops up only for a fraction of the  time and it can be annoying. It's also annoying that the loading wheel pops up :) I'd personally prefer seeing the dialog but this might not be the case for everyone. 

One more to add, for the screen-recording above, I used a very small backbone and submitted very short prompts. With bigger models and longer prompts, it already takes more than a couple of seconds to return the answer based on my experience (trying it on 3xRTX 6000 in our machines)

If you all think the loading wheel is better than the dialog for the chat, I can remove it 👍

fatihozturkh2o: >what is the sleep for?

It's just to make sure we see something when busy_dialog is called. I actually copy pasted it from our MV repo :D

psinger: I would stick to the loading wheel for now for the chat also.

fatihozturkh2o: >I would stick to the loading wheel for now for the chat also.

Done: https://github.com/h2oai/h2o-llmstudio/pull/49/commits/e68f1444b760a917dc13a82bce58480c27701a28

pascal-pfeiffer (approved):
lgmt!

thanks again @fatihozturkh2o

PR #52, opened 2023-04-26
Fix copy config

maxjeblick:
Hotfix for bf1cbf6bac896f292489feadc569162d68a82363.

OpenAI secret and Neptune secret are not stored in config.yaml, so we need to set them as env variables before calling `train_wave.py ` and copy them over to the config again in `train.py`.

CLI users can:
- Specify the secrets in the config (.py/yaml), not recommended
- Use environment variables, these will be copied over to the config (but not saved in there).

### DIFF
--- a/llm_studio/src/utils/utils.py
+++ b/llm_studio/src/utils/utils.py
@@ -33,10 +33,8 @@ def set_environment(cfg):
     """Sets and checks environment settings"""
     if not cfg.environment.openai_api_token:
         cfg.environment.openai_api_token = os.getenv("OPENAI_API_KEY", "")
-        print("OPENAI_API_KEY", cfg.environment.openai_api_token)
     if not cfg.logging.neptune_api_token:
         cfg.logging.neptune_api_token = os.getenv("NEPTUNE_API_TOKEN", "")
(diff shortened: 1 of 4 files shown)


llm_studio/src/utils/utils.py
```diff
@@ -31,7 +31,10 @@ def set_seed(seed: int = 1234) -> None:
 
 def set_environment(cfg):
     """Sets and checks environment settings"""
-
+    if not cfg.environment.openai_api_token:
+        cfg.environment.openai_api_token = os.getenv("OPENAI_API_KEY", "")
+    if not cfg.logging.neptune_api_token:
+        cfg.logging.neptune_api_token = os.getenv("NEPTUNE_API_TOKEN", "")
     os.environ["OPENAI_API_KEY"] = cfg.environment.openai_api_token
```
pascal-pfeiffer, line 38: i guess, this line is not needed as we are using the key with the line below. But doesn't need to be addressed here.

pascal-pfeiffer (approved):
Thank you for the hotfix!

Neptune and openai keys are being correctly passed and they are not saved to any yaml config files as expected.

PR #54, opened 2023-04-26
Filter out empty stop tokens

maxjeblick:
Filters out empty stop tokens (`""`) that will trigger StoppingCriteria at first token.


pascal-pfeiffer (approved):
thanks for the quick fix! 

lgtm

PR #59, opened 2023-04-27
hf push

pascal-pfeiffer:
- add yaml cfg
- add modelcard
- add pipeline


model_card_template.md
```diff
@@ -0,0 +1,84 @@
...
+pip install accelerate==0.18.0
+```
+
+```python
+import torch
+from transformers import pipeline
+
+generate_text = pipeline(model="{{repo_id}}", torch_dtype=torch.bfloat16, trust_remote_code=True, device_map="auto")
+
+res = generate_text("Why is drinking water so healthy?", max_new_tokens=100)
```
maxjeblick, line 34: You may want to add special tokens in the example prompt, see also https://github.com/h2oai/h2o-llmstudio/issues/61

pascal-pfeiffer: What do you mean?

pascal-pfeiffer: But it will be more difficult for the chained prompts. Maybe, I need to rework it again.

pascal-pfeiffer: But it will be more difficult for the chained prompts. Maybe, I need to rework it again.

psinger: I think we can ignore chained here, this should be rather clear.

app_utils/sections/experiment.py
```diff
@@ -1614,17 +1615,83 @@ async def experiment_push_to_huggingface_dialog(q: Q, error: str = ""):
...
-            token=q.client["experiment/display/push_to_huggingface/api_key"],
+            repo_id=repo_id,
             private=True,
         )
 
+        # push model card to hub
+        card_data = huggingface_hub.ModelCardData(
+            language="en",
+            library_name="transformers",
+            tags=["gpt", "llm", "large language model"],
```
psinger, line 1626: Add h2o-llmstudio tag?

model_card_template.md
psinger: shall we potentially add the inference parameters from the experiment there?

app_utils/sections/experiment.py
```diff
@@ -1607,17 +1608,83 @@ async def experiment_push_to_huggingface_dialog(q: Q, error: str = ""):
...
+
+        # push pipeline to hub
+        template_env = Environment(loader=FileSystemLoader(searchpath="llm_studio/src/"))
+        pipeline_template = template_env.get_template("h2oai_pipeline_template.py")
+
+        data = {
+            "text_prompt_start": cfg.dataset.text_prompt_start,
+            "text_answer_separator": cfg.dataset.text_answer_separator,
+        }
+        if cfg.dataset.add_eos_token_to_prompt:
```
psinger, line 1673: will it work if this is not set?

model_card_template.md
```diff
@@ -0,0 +1,84 @@
...
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+tokenizer = AutoTokenizer.from_pretrained("{{repo_id}}", padding_side="left")
+model = AutoModelForCausalLM.from_pretrained("{{repo_id}}", torch_dtype=torch.bfloat16, device_map="auto")
+generate_text = H2OTextGenerationPipeline(model=model, tokenizer=tokenizer)
+
+res = generate_text("Why is drinking water so healthy?", max_new_tokens=100)
+print(res[0]["generated_text"])
+```
+
```
psinger, line 52: I think it would still be nice if we clearly state the syntax of the prompt here somewhere. Not everyone will look into the pipeline file to understand it.

psinger (changes requested):

model_card_template.md
pascal-pfeiffer: good idea, will add this

maxjeblick: `What do you mean? EOS, prompt start, answer start? These are added in the pipeline.`
Ok, I missed that

psinger (approved):

PR #62, opened 2023-04-27
Update readme

### LINKED ISSUE #61: [CODE IMPROVEMENT] Improve README for Huggingface integration

Add example code on how to use Huggingface models, in particular, how tokenization looks

`inputs = tokenizer("<|prompt|>How are you?<|endoftext|><|answer|>", return_tensors="pt", add_special_tokens=False).to("cuda")`

### Motivation

People may easily miss the correct  prompt template.

maxjeblick:
Fixes #61.

For some models, example prompts where not displayed correctly in the UI logs, e.g. if `EOS token == </s>`.
I've fixed it for the UI, terminal logs now show html formatted prompts. Not sure if there's an easy way to both get good output both in UI and in the terminal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ During the experiment, you can monitor the training progress and model performan
 ### Push to Hugging Face 🤗
 If you want to publish your model, you can export it with a single click to the [Hugging Face Hub](https://huggingface.co/)
 and share it with the community. To be able to push your model to the Hub, you need to have an API token with write access.
-You can also click the _Download model_ button to the model locally in Hugging Face format and use it in your own application.
+You can also click the **Download model** button to the model locally in Hugging Face format and use it in your own application.
 To use a converted model, you can use the following code snippet:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -132,6 +132,35 @@ During the experiment, you can monitor the training progress and model performan
 ### Push to Hugging Face 🤗
 If you want to publish your model, you can export it with a single click to the [Hugging Face Hub](https://huggingface.co/)
 and share it with the community. To be able to push your model to the Hub, you need to have an API token with write access.
+You can also click the **Download model** button to download the model locally.
```
psinger, line 135: Let's align this messaging with https://github.com/h2oai/h2o-llmstudio/pull/59

pascal-pfeiffer (approved):
lgtm

We should probably include the stopping logic in a follow-up. Same will need to be done for the huggingface model card.

Issue #60, opened 2023-04-27
[BUG] New Experiment resets API keys


psinger:
### 🐛 Bug

Starting a new experiment via New Experiment from an existing one, appears to reset the secret tokens.

![image](https://user-images.githubusercontent.com/1677826/234817362-c6e43f74-f701-4bc9-8e56-0d4e5eec9a78.png)

Reason is most likely that new YAML functionality does not preserve the tokens, so they would need to be reloaded from default settings.

psinger:
Use previous experiment weights

seems also broken

PR #68, opened 2023-04-27
Config fixes

psinger:
Closes https://github.com/h2oai/h2o-llmstudio/issues/60

Also removes api keys from cfg and moves them to pure env variables

@maxjeblick please make sure the visibility changes I made are OK - they are needed to fix `pretrained_weights` stuff

### DIFF
--- a/app_utils/utils.py
+++ b/app_utils/utils.py
@@ -11,13 +11,13 @@ import shutil
 import socket
 import subprocess
 import time
+import uuid
 import zipfile
 from collections import defaultdict
 from contextlib import closing
 from functools import partial
-from typing import Any, DefaultDict, Dict, List, Optional, Tuple, Type, Union
+from typing import Any, DefaultDict, List, Optional, Tuple, Type, Union
 
-import dill
 import GPUtil
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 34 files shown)


app_utils/utils.py
```diff
@@ -1571,9 +1571,10 @@ def start_experiment(cfg: Any, q: Q, pre: str, gpu_list: Optional[List] = None)
     process_queue = list(set(all_process_queue))
 
     secrets = {
-        "NEPTUNE_API_TOKEN": cfg.logging.neptune_api_token,
-        "OPENAI_API_KEY": cfg.environment.openai_api_token,
+        "NEPTUNE_API_TOKEN": q.client["default_neptune_api_token"],
+        "OPENAI_API_KEY": q.client["default_openai_api_token"]
     }
+    print(secrets)
```
maxjeblick, line 1577: Artifact from debugging?

llm_studio/src/utils/utils.py
```diff
@@ -31,16 +31,11 @@ def set_seed(seed: int = 1234) -> None:
 
 def set_environment(cfg):
     """Sets and checks environment settings"""
-    if not cfg.environment.openai_api_token:
-        cfg.environment.openai_api_token = os.getenv("OPENAI_API_KEY", "")
-    if not cfg.logging.neptune_api_token:
-        cfg.logging.neptune_api_token = os.getenv("NEPTUNE_API_TOKEN", "")
-    os.environ["OPENAI_API_KEY"] = cfg.environment.openai_api_token
-    openai.api_key = cfg.environment.openai_api_token
```
maxjeblick, line 39: This line is no longer needed?

llm_studio/src/utils/config_utils.py
```diff
@@ -106,7 +106,7 @@ def convert_cfg_to_nested_dictionary(cfg: ConfigProblemBase) -> dict:
 
     for k, v in cfg_dict.items():
 
-        if k.startswith("_") or cfg._get_visibility(k) < 0:
```
maxjeblick, line 109: What's the rationale behind removing `cfg._get_visibility(k) < 0`?

With this change, yaml config will also save non visible configuration items that are of type `KNOWN_TYPE_ANNOTATIONS`. 
I don't think it's necessarily a disadvantage to also store those fields, as they are part of the configuration.

psinger: It is needed becasue we set `pretrained_weights` when activating `Use previous experiment weights` but it gets visibility -1 so the yaml does not store it. We can either add it explicitly here as another exception, but I thought it should not hurt to also store those not being visible.
wdyt?

llm_studio/src/utils/utils.py
psinger: We are setting it here, or?
`openai.api_key = os.getenv("OPENAI_API_KEY", "")`

maxjeblick: Ah yes missed the lower part

llm_studio/src/utils/config_utils.py
maxjeblick: Yes, I think it's good to store all configurations, as visibility options may change.

maxjeblick (approved):
LGTM thanks!

PR #72, opened 2023-04-28
Closes #70

### LINKED ISSUE #70: [CODE IMPROVEMENT] Use cache with GC
When using gradient checkpointing this message gets printed a lot:
`use_cache=True is incompatible with gradient checkpointing. Setting use_cache=False...`

We are already disabling it though:
https://github.com/h2oai/h2o-llmstudio/blob/main/llm_studio/src/utils/modeling_utils.py#L648

Need to investigate why this message still pops up.

psinger:
Closes https://github.com/h2oai/h2o-llmstudio/issues/70

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,12 @@ isort: pipenv
 black: pipenv
 	$(PIPENV) run black .
 
+.PHONY: test
+test: reports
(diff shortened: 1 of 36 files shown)


maxjeblick (approved):
LGTM

Issue #71, opened 2023-04-28
[CODE IMPROVEMENT] Exponential backoff for API rate limits


psinger:
https://platform.openai.com/docs/guides/rate-limits/error-mitigation

PR #80, opened 2023-05-02
Improve gpt metric calculation

psinger:
This PR adds

exponential backoff
reduced nr of workers

for gpt-based metric calculation

The backoff is less important than expected, mostly improvements come from reducing the nr of workers.

closes https://github.com/h2oai/h2o-llmstudio/issues/71


pascal-pfeiffer (approved):
lgtm

PR #81, opened 2023-05-02
default gpt eval max

### LINKED ISSUE #45: [FEATURE] GPT_EVAL_MAX editable via UI
### 🚀 Feature

Soweit ich sehen konnte, gibt es derzeit kein Bedienfeld, mit dem ich die angesprochenen Environment Variable editieren kann, was dazu führt, dass das Statement WARNING: `More than 100 validation records. Safeguarding against OpenAI API costs. Setting metric to BLEU. Change GPT_EVAL_MAX to run GPT` im Log "verschluckt" wird.

### Motivation

If possible, it would probably make sense not to have to reboot the system to set this value. It might be unlikely that several experiments run in parallel - nevertheless, this could probably be avoided.

What do you think?

pascal-pfeiffer:
closes #45


maxjeblick (approved):
LGTM, works as expected

Issue #41, opened 2023-04-25
[FEATURE] Add more inference parameters


psinger:
### 🚀 Feature

Add more inference parameters:

- Top P
- Top K

maxjeblick:
Fixed by #85

PR #85, opened 2023-05-03
Max/chat block

maxjeblick:
This PR 

- blocks the chat window if an experiment is currently running on GPU 0
- adds additional inference parameters for the chat tab

The chat window itself is not blocking, i.e. in theory, a queued experiment could start while the user is using the chat window. While mutual blocking is probably the best solution, I'm not sure if it's currently worth the effort to implement.

Screenshot 2023-05-03 at 10-46-39 H2O LLM Studio


pascal-pfeiffer:
Very nice! Can we wrap it in a `ui.expander` element to not block the view on smaller screens?

maxjeblick:
> Very nice! Can we wrap it in a ui.expander element to not block the view on smaller screens?

Good point, I added it.

app_utils/sections/experiment.py
```diff
@@ -1729,8 +1796,7 @@ def load_cfg_model_tokenizer(experiment_path, merge=False, device="cuda:0"):
     cfg.architecture.gradient_checkpointing = False
     cfg.environment._device = device
     cfg.environment._local_rank = 0
-
-    cfg.prediction.num_history = 2
+    cfg.prediction._num_history = 2
```
psinger, line 1799: Could we also expose that in the UI?

psinger (changes requested):
Thanks @maxjeblick - great features.

Here is some more feedback:

- Could we add an additional "Clear chat" / "New Chat" button that replicates the `--clear `functionality?
- Maybe could be useful to directly add these two inference parameters then also for this pr: https://github.com/h2oai/h2o-llmstudio/issues/41
- The layout is not idea, but I know that it is hard. I gues it wont be possible to dynamically expand the space when it is collated vs not collated, right?
- Shall we safeguard the model loading a bit more? Maybe checking for available memory? Maybe giving an option for device? Lets open maybe a new issue for that.

psinger:
Sorry, but actually one more thing we should add, markdown support for chatbot:
https://github.com/h2oai/wave/issues/1927

Will need new nightly wave I assume.

maxjeblick:
> Could we add an additional "Clear chat" / "New Chat" button that replicates the --clear functionality?

I have added a button and removed the `--clear `code.

> Maybe could be useful to directly add these two inference parameters then also for this issue:
> https://github.com/h2oai/h2o-llmstudio/issues/41

I have added those. Default is `top_k = 0` and `top_p = 1`, with `top_k = 0` disabling top k sampling.

> The layout is not ideal, but I know that it is hard. I gues it wont be possible to dynamically expand the space when it is collated vs not collated, right?

I haven't found an easy way to dynamically expand the space.

> Shall we safeguard the model loading a bit more? Maybe checking for available memory? Maybe giving an option for device? Lets open maybe a new issue for that.

Yes, let's maybe address this in a subsequent PR.


> Sorry, but actually one more thing we should add, markdown support for chatbot:
> https://github.com/h2oai/wave/issues/1927

Should we address this in this PR or open a subsequent issue?

psinger (approved):
Thanks, looks cool!

Just minor things, feel free to skip them:
- Can we put the new inference parameters in a better order before `Stop Tokens` in the start experiment screen?
- When Im clearing the chat, it also collapses the expert settings again. Actually I would even suggest to have it uncollapsed by default to directly see it always.

For this issue I have an idea: https://github.com/h2oai/h2o-llmstudio/issues/101

We could always take the first bubble as kind-of the user request, so something like:
Load experiment blabla
Clear the chat

And then the second bubble on the left is the one from the bot "model loaded how can I help you" and one can continue from there.

Can do that in a separate PR. Same as markdown.

PR #97, opened 2023-05-05
Disable adapter for zero shot eval

### LINKED ISSUE #96: [CODE IMPROVEMENT] Disable adapter for zero shot eval

Disable adapter for zero shot eval

### Motivation

Adapter is randomly initialized at that stage.

pascal-pfeiffer:
closes #96

fixes peft to 0.3.0 that now includes everything that we needed from the dev version.


psinger:
Why do we need this explicitly?

The LORA weights are initialized with zero, so should not impact the backbone.

Issue #75, opened 2023-04-28
[FEATURE] Select device that should be used to merge weights (for Download and for push to Hugging Face)


pascal-pfeiffer:
### 🚀 Feature

Select device that should be used to merge weights (for Download and for push to Hugging Face)

### Motivation

CPU might be slow for larger models

PR #110, opened 2023-05-12
HF Push and Download improvements

psinger:
This PR adds two new options when pushing to HF: device and account name

For downloading of weights, it just manually checks if experiments are running and in that case switches to CPU.

Closes https://github.com/h2oai/h2o-llmstudio/issues/75

### DIFF
--- a/app_utils/sections/experiment.py
+++ b/app_utils/sections/experiment.py
@@ -1577,7 +1577,6 @@ async def experiment_download_model(q: Q, error: str = ""):
 
 async def experiment_push_to_huggingface_dialog(q: Q, error: str = ""):
     if q.args["experiment/display/push_to_huggingface"] or error:
-
         devices = ["cpu"] + [f"cuda:{idx}" for idx in range(torch.cuda.device_count())]
         default_device = "cuda:0"
 
@@ -1591,7 +1590,7 @@ async def experiment_push_to_huggingface_dialog(q: Q, error: str = ""):
         try:
             huggingface_hub.login(q.client["default_huggingface_api_token"])
             user_id = huggingface_hub.whoami()["name"]
-        except:
+        except Exception:
             user_id = ""
 
         dialog_items = [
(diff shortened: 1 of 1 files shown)


app_utils/sections/experiment.py
```diff
@@ -1596,19 +1649,23 @@ async def experiment_push_to_huggingface_dialog(q: Q, error: str = ""):
...
-        huggingface_hub.login(
-            q.client["experiment/display/push_to_huggingface/api_key"]
-        )
-        user_id = huggingface_hub.whoami()["name"]
+        user_id = q.client["experiment/display/push_to_huggingface/account_name"]
+        if user_id == "":
+            huggingface_hub.login(
+                q.client["experiment/display/push_to_huggingface/api_key"]
+            )
+            user_id = huggingface_hub.whoami()["name"]
```
pascal-pfeiffer, line 1668: If no default token is provided (settings page) but a `user_id` in the textbox the user will never be logged in

app_utils/sections/experiment.py
```diff
@@ -1565,14 +1577,54 @@ async def experiment_download_model(q: Q, error: str = ""):
...
+            ui.textbox(
+                name="experiment/display/push_to_huggingface/account_name",
+                label="Account Name",
+                value=user_id,
+                width="500px",
+                required=False,
+                tooltip=(
+                    "The account name on HF to push the model to. "
+                    "Leaving it empty will push it to the default user account.",
+                ),
```
pascal-pfeiffer, line 1607: that's a tuple, rm trailing comma

pascal-pfeiffer (changes requested):
Thanks a lot, that makes the weights merging much faster.

Two minor issue that I am still seeing.

Additionally, it would make sense to show a better error when the user credentials are wrong.

pascal-pfeiffer (approved):
lgtm

PR #111, opened 2023-05-14
Creating the option to download checkpoints as models

pizzato:
Something like this will enable people to run large jobs and download models mid-way, as well as not to worry about stopping a large job and not getting any model out to test.

Enables a "Download Model from checkpoint" and "Push checkpoint to huggingface" after an experiment generated a checkpoint file.

image

### DIFF
--- a/app_utils/sections/experiment.py
+++ b/app_utils/sections/experiment.py
@@ -1095,6 +1095,25 @@ async def experiment_display(q: Q) -> None:
         )
     ]
 
+    if status != "finished":
+        buttons += [
+            ui.button(
+                name="experiment/display/download_model",
+                label="Donwload checkpoint",
+                primary=False,
+                disabled=False,
+                tooltip=None,
+            ),
+            ui.button(
+                name="experiment/display/push_to_huggingface",
+                label="Push checkpoint to huggingface",
+                primary=False,
+                disabled=False,
+                tooltip=None,
+            ),
+        ]
+
+
     if status == "finished":
         buttons += [
             ui.button(
--- a/app_utils/sections/experiment.py
+++ b/app_utils/sections/experiment.py
@@ -1096,22 +1096,25 @@ async def experiment_display(q: Q) -> None:
     ]
 
     if status != "finished":
-        buttons += [
-            ui.button(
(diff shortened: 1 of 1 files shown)


pascal-pfeiffer:
Thank you for your contribution @pizzato 
I like the idea, but also see a few issues that come with it:
- I assume the provided code would fail if the checkpoint is overwritten during upload.
- Currently, the checkpoints are only saved at the end of the training loop or if `save_best_checkpoint` is enabled.
- It would be hard to tell which checkpoint was downloaded / uploaded to HF. 

As this may introduce confusing behaviour I am hesitant to add it to the core functionality. One can always go ahead and pull the files from the filesystem, too.

pizzato:
Hi @pascal-pfeiffer, on  your points:

> I assume the provided code would fail if the checkpoint is overwritten during upload.

I agree that HF upload will certainly cause trouble as it takes a very long time. This should be removed

> Currently, the checkpoints are only saved at the end of the training loop or if save_best_checkpoint is enabled.
   
Correct, hence the test for the button only to appear when the checkpoint file is there. 

> It would be hard to tell which checkpoint was downloaded / uploaded to HF.

True, unless I missed something, it also seems that only the best checkpoint is recorded. So if a checkpoint performs worse then the previous, it isn't recorded, and there is no indication of which checkpoint it is. It would be good to add that to the tool.

I feel the functionality to extract a working model from a checkpoint is useful. I think the best solution here would be to add an identification of the checkpoint with the epoch followed by a download checkpoint model which carries that id forward. 

I can close this pull request and potentially re-submit (if wanted) a similar request when the epoch is added to the checkpoint. 



Happy to modify and suggest something, unless it won't fall into the tool overall design.

pascal-pfeiffer:
I still like the idea a lot to download models that are not finished. "Just" needs a good UX to it, too. 

Maybe the better idea is to add a pause option for running experiments together with semi-frequent checkpointing including the optimizer state. This would be need to set in advance, and could be worthful for long running experiments. The downside is the required large disk space. What do you think @pizzato ?

If you are willing to work on such a feature we would be more than happy about a PR.

PR #113, opened 2023-05-16
update version

maxjeblick:
Bump Version 0.0.2

Also includes a small fix that caused the chat setting not to be cleared in the UI in case of OOM when loading the model.


pascal-pfeiffer (approved):
lgtm

thanks for the fix

PR #114, opened 2023-05-16
update readme to include support for installing nvidia drivers if req…

jfarland:
This PR simply updates the readme to provide more support for installing nvidia drivers if the OS / machine image isn't preconfigured to support.


pascal-pfeiffer (approved):
Thank you for providing this little guide! Lgtm

PR #121, opened 2023-05-22
chat bubbles switched, user on right side

### LINKED ISSUE #101: [FEATURE] Switch the bubbles on the Model Chat
### 🚀 Feature

Every texting application has the user chats on the right side and the responding bubbles on the left. Can we switch this in LLM chat?

pascal-pfeiffer:
image

To also switch the colors, we probably have to wait for a wave fix

Updated wave to always take the latest nightly, newlines are now supported in the chat.

closes #101

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,14 +14,17 @@ pipenv:
 .PHONY: setup
 setup: pipenv
 	$(PIPENV) install --verbose --python $(PYTHON_VERSION)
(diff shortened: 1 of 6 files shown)


pascal-pfeiffer:
I don't see the issue on my machine and I don't see a way how we can control it here. So should probably be addressed at wave component level.

PR #131, opened 2023-06-01
Max/4bit

maxjeblick:
This PR adds 4bit training/inference.
I have tested with `EleutherAI/gpt-neox-20b` where 8 bit yields OOM. I have also compared `EleutherAI/gpt-j-6B` 4bit/8bit training and its corresponding chat responses.

There is an issue with merging LORA back which can yield to OOM issues: https://github.com/h2oai/h2o-llmstudio/discussions/130  I'll have a look and see if it can be fixed and added to this PR.

I'm also not sure about this line: https://github.com/h2oai/h2o-llmstudio/blob/main/llm_studio/src/utils/modeling_utils.py#L72 
It's currently left as is, I haven't found any issues with it so far.


psinger:
> I'm also not sure about this line: https://github.com/h2oai/h2o-llmstudio/blob/main/llm_studio/src/utils/modeling_utils.py#L72
It's currently left as is, I haven't found any issues with it so far.

IIRC this was needed for training and loading weights in different precisions

app_utils/sections/experiment.py
```diff
@@ -1595,17 +1595,17 @@ async def experiment_download_model(q: Q, error: str = ""):
...
         )
-        if num_running_queued > 0:
-            logger.info(
-                "Preparing model on CPU as there are experiments running. "
-                "This might slow down the progress."
-            )
+        if num_running_queued > 0 or (
+            cfg.training.lora and cfg.architecture.backbone_dtype in ("int4", "int8")
+        ):
+            logger.info("Preparing model on CPU. " "This might slow down the progress.")
```
pascal-pfeiffer, line 1608: format

llm_studio/python_configs/text_causal_language_modeling_config.py
```diff
@@ -228,7 +228,8 @@ def __post_init__(self):
         super().__post_init__()
 
         self._possible_values["backbone_dtype"] = possible_values.String(
-            values=("float16", "bfloat16", "int8", "float32"), allow_custom=False
+            values=("float16", "bfloat16", "int8", "int4", "float32"),
```
pascal-pfeiffer, line 231: can we order that by memory consumption maybe?

pascal-pfeiffer:
and let's also add some words in the README.md about 4bit training

llm_studio/src/utils/modeling_utils.py
```diff
@@ -504,7 +499,9 @@ def create_nlp_backbone(cfg, model_class=AutoModel, kwargs={}) -> Any:
         backbone.resize_token_embeddings(cfg.tokenizer._vocab_length)
 
     if cfg.training.lora:
-        backbone = prepare_model_for_lora_training(backbone, layer_norm_names=[])
+        backbone = prepare_model_for_kbit_training(
+            backbone, use_gradient_checkpointing=cfg.architecture.gradient_checkpointing
```
psinger, line 503: We are doing the gradient checkpointing code already outside. So this needs to be set to False.

See the old function `prepare_model_for_lora_training`.

pascal-pfeiffer: It shouldn't hurt, right?

If so, I would leave it like this to avoid confusion then

psinger: It is redundant and might lead to issues. I would just set it to False there.

psinger:
One more note:
Inference in 4bit is extremely slow for me.

Would it maybe make sense to force 8bit in Chat window?
Or even better, make it somehow an option?
Might be a separate issue.

psinger:
Found this:
https://github.com/huggingface/peft/blob/main/examples/fp4_finetuning/finetune_fp4_opt_bnb_peft.py

Here they are also setting `llm_int8_threshold` - does it have an impact in 4bit?

maxjeblick:
> Here they are also setting llm_int8_threshold - does it have an impact in 4bit?

Should only affect 8bit:
https://github.com/huggingface/transformers/blob/main/src/transformers/utils/bitsandbytes.py#L133

psinger:
Not sure this was lucky or has impact, but I got better inference speed with these settings:

```
elif cfg.architecture.backbone_dtype == "int4":
        kwargs["device_map"] = {"": cfg.environment._device}
        quantization_config = BitsAndBytesConfig(
            load_in_4bit=True,
            bnb_4bit_compute_dtype=torch.float16,
            llm_int8_has_fp16_weight=True,
            bnb_4bit_use_double_quant=True,
            bnb_4bit_quant_type="nf4",
            llm_int8_threshold=0.0
        )
        # need to force pretrained
        cfg.architecture.pretrained = True
        kwargs["torch_dtype"] = torch.float16
```

psinger:
Let's add this also to int8:

`kwargs["torch_dtype"] = torch.float16`

And do you think `bnb_4bit_use_double_quant=True` is useful?

maxjeblick:
> Getting errors when training in 4bit, and then doing inference on float16.

Is this issue still present? I started a new experiment in fp16, using previous 4bit weights; that works.

psinger:
> > Getting errors when training in 4bit, and then doing inference on float16.
> 
> Is this issue still present? I started a new experiment in fp16, using previous 4bit weights; that works.

I think after you made the changes above it works now.

psinger:
@maxjeblick did you try pushing to HF and reloading?

psinger:
Cool, if possible lets also try the HF push.

psinger (approved):
Thanks!

maxjeblick:
Checked HF hub, it also works.

PR #136, opened 2023-06-05
gpt eval with 4096 toks and temp 0.0

pascal-pfeiffer:


psinger (approved):

pascal-pfeiffer:
1024 is the completion output

PR #142, opened 2023-06-07
Manual loss & chained setting

psinger:
This PR addresses to points:

First, it introduces a new setting `limit_chained_samples` which limits the chained samples to the longest one. Please feel free to suggest a better name for that.

Second, it reworks loss calculation by adding functionality to calculate the loss in our code base instead of relying on the one in `transformers` model class. This will also allow us long-term more flexibility, in terms of adding new loss functions, weighted loss, etc.

For now, there are two BCE loss variations: 
- `TokenCrossEntropy` calculates the loss as before on token level
- `SampleCrossEntropy` calculates the loss separate per sample in a batch, and then averages per sample

Experiments seem to indiciate that in generel the old way works better.

### DIFF
--- a/llm_studio/src/models/text_causal_language_modeling_model.py
+++ b/llm_studio/src/models/text_causal_language_modeling_model.py
@@ -195,7 +195,6 @@ class Model(nn.Module):
             output = self.backbone(
                 input_ids=batch["input_ids"],
                 attention_mask=batch["attention_mask"],
-                labels=None,
             )
 
             if calculate_loss:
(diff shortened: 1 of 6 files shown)


llm_studio/src/models/text_causal_language_modeling_model.py
```diff
@@ -193,11 +195,12 @@ def forward(
             output = self.backbone(
                 input_ids=batch["input_ids"],
                 attention_mask=batch["attention_mask"],
-                labels=batch["labels"],
+                labels=None,
```
pascal-pfeiffer, line 198: or remove line entirely to avoid confusion? None is the default, right?

pascal-pfeiffer (approved):
Thank you, lgtm!

Two comments in the code. Your choice to address them or merge as is.

PR #152, opened 2023-06-11
Pp/rlhf

pascal-pfeiffer:
This PR adds RLHF as an option for fine-tuning a LLM.
It makes heavy use of the awesome [trl](https://github.com/lvwerra/trl) lib.

- [x] RLHF on single GPU setups
- [x] shared weights between reference model and active model to reduce the huge memory requirement when having three models running
- [x] DDP support
- [ ] Train your own reward model

Some logs to show that it is possible to overfit the reward model (though it doesn't translate to any increase in GPT score, yet):

Reward:
train_ppo_mean_scores

GPT3.5 score:
validation_GPT3 5

Validation loss:
validation_loss

[internal logs] https://app.neptune.ai/o/Zoo/org/h2o-llm/runs/details?viewId=99593e71-d48c-4a30-827b-b5efe805f0de&detailsTab=charts&shortId=HOL-1534


maxjeblick (commented):
Thanks a lot for this awesome new feature!

I'm currently still going trough the code (looks good so far!), some preliminary feedback:

- I'm getting WARNING: `use_cache=True` is incompatible with gradient check pointing. Setting `use_cache=False` warnings throughout the training (default settings with `h2oai/h2ogpt-gm-oasst1-en-2048-falcon-7b-v2`).
- We probably should fail ddp training as it is not yet supported.
- There are some parameters like gradient accumulation that are not used for rlhf. Would it be possible to hide those when rlhf is enabled (never checked if it's possible with nesting)?

llm_studio/src/datasets/text_causal_language_modeling_ds.py
```diff
@@ -172,15 +177,17 @@ def get_validation_collate_fn(self):
...
-    ) -> Dict:
+    def postprocess_batch_predictions(self, cfg: Any, output: Dict) -> Dict:
         predicted_text = [
             self.tokenizer.decode(ids, skip_special_tokens=True).strip()
             for ids in output["predicted_answer_ids"]
         ]
         output["predicted_text"] = np.array(predicted_text)
-        del output["predicted_answer_ids"]
+
+        if not cfg.training.use_rlhf:
```
maxjeblick, line 187: Do you think it makes sense to repopulate `predicted_answer_ids` in `train.py` instead?

train.py
```diff
@@ -387,7 +480,7 @@ def run_train(
     if cfg.environment._distributed:
         torch.distributed.barrier()
 
-    return val_data, val_loss, val_metric, batch
```
maxjeblick, line 390: Nice refactoring! In line 203, `batch = None` can be deleted, as well.

train.py
```diff
@@ -500,12 +594,24 @@ def run(cfg: Any) -> None:
...
+            reward_model = None
+
         # load model weights
         if cfg.architecture.pretrained_weights != "":
             # Do not load strictly if continue training from the previous experiment
             load_checkpoint(cfg, model, strict=cfg.training.epochs == -1)
 
     model.to(cfg.environment._device)
+    if cfg.training.use_rlhf:
+        if cfg.training.offload_reward_model:
```
maxjeblick, line 611: Like the offloading!

llm_studio/src/trl/trainer.py
```diff
@@ -0,0 +1,851 @@
...
+
+        model_inputs = self.prepare_model_inputs(queries, responses)
+
+        model_inputs_names = list(model_inputs.keys())
+
+        with torch.no_grad():
+            all_logprobs, _, values, masks = self.batched_forward_pass(
+                self.model, queries, responses, model_inputs
+            )
+            with self.model.backbone.disable_adapter():
```
maxjeblick, line 340: Awesome method to avoid having the reference model as a separate model.
We probably should fail if LORA is disabled or force enable lora with a warning in the logs.

llm_studio/src/datasets/text_causal_language_modeling_ds.py
```diff
@@ -322,12 +339,18 @@ def _read_data(self, idx: int, sample: Dict) -> Dict:
...
 
         input_ids = torch.cat([torch.cat(sample) for sample in samples])
-        prompt_mask = torch.cat(
-            [
-                torch.cat([torch.ones_like(sample[0]), torch.zeros_like(sample[1])])
-                for sample in samples
-            ]
-        ).to(torch.bool)
+
+        if self.cfg.training.use_rlhf and self.mode == "train":
```
maxjeblick, line 343: The `_read_data` method is a bit hard to follow. But I can also refactor after this PR has been merged.

llm_studio/src/datasets/text_causal_language_modeling_ds.py
```diff
@@ -294,12 +301,17 @@ def _get_sample(self, idx):
                 ],
                 dim=0,
             )
-        return [prompt_encodings, answer_encodings]
+
+        if self.cfg.training.use_rlhf and self.mode == "train":
+            return [prompt_encodings]
+        else:
+            return [prompt_encodings, answer_encodings]
```
maxjeblick, line 308: Answer encodings would still be needed for the chat history (when using `parent_id_column`)?

llm_studio/src/trl/trainer.py
```diff
@@ -0,0 +1,851 @@
...
+def logprobs_from_logits(logits, labels):
+    """
+    See: https://github.com/pytorch/pytorch/issues/563#issuecomment-330103591
+    """
+    logp = F.log_softmax(logits, dim=2)
+    logpy = torch.gather(logp, 2, labels.unsqueeze(2)).squeeze(-1)
+    return logpy
+
+
+def whiten(values, shift_mean=True):
```
maxjeblick, line 78: Seems to be unused

llm_studio/src/models/text_causal_language_modeling_model.py
```diff
@@ -142,66 +220,259 @@ def generate(self, batch: Dict, cfg: Any):
...
+            )
 
+        if not self.training or generate:
+            outputs["predicted_answer_ids"] = (
+                self.generate(batch, self.cfg).detach().cpu()
+            )
         return outputs
+
+
+class GPTNeoXRewardModelConfig(GPTNeoXConfig):
```
maxjeblick, line 333: Could we move the reward models to a separate file?

llm_studio/python_configs/text_causal_language_modeling_config.py
```diff
@@ -196,6 +234,24 @@ def __post_init__(self):
...
+                "initial_kl_coefficient",
+                "kl_target",
+                "kl_horizon",
+                "advantages_gamma",
+                "advantages_lambda",
+                "ppo_clip_policy",
+                "ppo_clip_value",
+                "scaling_factor_value_loss",
+                "ppo_epochs",
+                "ppo_batch_size",
```
maxjeblick, line 251: `ppo_generate_temperature` is missing here

llm_studio/src/models/text_causal_language_modeling_model.py
```diff
@@ -142,66 +220,259 @@ def generate(self, batch: Dict, cfg: Any):
...
 
         return output
 
     def forward(
         self,
         batch: Dict,
-        calculate_loss: bool = True,
+        generate: bool = False,
+        padding: bool = True,
+        is_ref_model: bool = False,
```
maxjeblick, line 276: I think one can remove `is_ref_model` flag and also compute loss/value in the reference model (they will not be used afterwards, of course). This would result in slightly slimmer code. WDYT?

llm_studio/src/datasets/text_causal_language_modeling_ds.py
```diff
@@ -322,12 +339,18 @@ def _read_data(self, idx: int, sample: Dict) -> Dict:
...
+        else:
+            prompt_mask = torch.cat(
+                [
+                    torch.cat([torch.ones_like(sample[0]), torch.zeros_like(sample[1])])
+                    for sample in samples
+                ]
+            ).to(torch.bool)
         attention_mask = torch.ones_like(input_ids)
 
         labels = input_ids.clone()
```
maxjeblick, line 356: We could think about skipping label creation for rlhf (I should not be needed for rlhf).

llm_studio/src/trl/trainer.py
maxjeblick: Also, force embedding gradients should be disabled.

train.py
```diff
@@ -259,59 +272,139 @@ def run_train(
...
-                    "NaN caught in loss during training. "
-                    "Please, reduce learning rate, change dtype, "
-                    "or disable mixed precision. "
-                    "Alternatively, gradient clipping may help to stabilize training."
+            if cfg.training.use_rlhf:
+                torch.inference_mode(mode=True)
+                logger.debug("Rollout: Generating response from active model")
+                output_dict = {}
+                output_dict["predicted_answer_ids"] = model.generate(
+                    batch, model.cfg, remove_prompt=True
```
maxjeblick, line 279: I noticed that I mixed up `max_length_answer` with min length and max length for inference (causing OOMs).

llm_studio/src/datasets/text_causal_language_modeling_ds.py
```diff
@@ -62,7 +62,9 @@ def __init__(self, df: pd.DataFrame, cfg: Any, mode: str = "train"):
 
         self.tokenizer = get_tokenizer(cfg)
 
-        self.prompts = get_texts(df, self.cfg, separator="")
+        self.raw_prompts = get_texts(df, self.cfg, separator="")
+        self.prompts = [self.parse_prompt(cfg, prompt) for prompt in self.raw_prompts]
```
maxjeblick, line 66: It is possible to move `.parse_prompt` to the `_get_sample` method and to the example logging.
By this, one does not need to keep two versions of the prompts.

llm_studio/src/datasets/text_causal_language_modeling_ds.py
pascal-pfeiffer: You mean making a copy of the `predicted_answer_ids` first and then add back to the dict after the post processing? sounds a bit complicated to me, or I am probably missing something.

llm_studio/src/datasets/text_causal_language_modeling_ds.py
pascal-pfeiffer: In this case, afaik, the "old" answers are in the prompt. But it's definetly tricky, as they need to be passed to the reward model, too. Which is not really trained on full conversations but rather single question, single answer if I see that correctly.

llm_studio/src/datasets/text_causal_language_modeling_ds.py
pascal-pfeiffer: we use `parse_prompt` in the chat

maxjeblick: Ah sorry, I mistyped my idea:
What I meant was to remove `self.prompts` (and rename `self.raw_prompts` to `self.prompts`) and call `self.parse_prompt` in the `_get_sample` method and when logging an example.

llm_studio/src/datasets/text_causal_language_modeling_ds.py
maxjeblick: Should be fine the way it is currently

llm_studio/python_configs/text_causal_language_modeling_config.py
```diff
@@ -158,16 +174,24 @@ def __post_init__(self):
...
         self._possible_values["differential_learning_rate"] = self._possible_values[
             "learning_rate"
         ]
+        self._possible_values["reward_model"] = possible_values.String(
+            values=(
+                "OpenAssistant/reward-model-deberta-v3-large-v2",
+                "OpenAssistant/oasst-rm-2.1-pythia-1.4b-epoch-2.5",
+                "OpenAssistant/oasst-rm-2-pythia-6.9b-epoch-1",
+            ),
+            allow_custom=True,
```
maxjeblick, line 190: `allow_custom=False` as `get_score` method currently only supports the models listed above.

llm_studio/src/models/text_causal_language_modeling_model.py
```diff
@@ -146,67 +214,114 @@ def generate(self, batch: Dict, cfg: Any):
...
         transformers_logging.set_verbosity(verbosity)
 
-        # Mask the prompt tokens
-        output[:, : batch["prompt_input_ids"].shape[1]] = pad_token_id
+        if remove_prompt:
+            # remove the prompt tokens
+            output = output[:, input_ids.shape[1] :]
+        else:
+            # Mask the prompt tokens
+            output[:, : input_ids.shape[1]] = pad_token_id
```
maxjeblick, line 261: Would it be possible to always remove the prompt?
We still mask it, and decoding in  `postprocess_batch_predictions` will skip pad tokens.

pascal-pfeiffer: yes, probably makes more sense. i didn't want to change the original behavior, but this looks reasonable.

llm_studio/src/datasets/text_causal_language_modeling_ds.py
```diff
@@ -345,6 +358,11 @@ def _read_data(self, idx: int, sample: Dict) -> Dict:
                 ):
                     break
                 samples.insert(0, self._get_sample(int(parent_idx)))
+                sample["reward_model_prompt_text"] = (
```
maxjeblick, line 361: As discussed, it should probably look something like
```
sample["reward_model_prompt_text"] = (
    self.raw_prompts[int(parent_idx)] + "<|endoftext|>" + self.answers[int(parent_idx)]
    + "<|endoftext|>"  # this is replaced later in the pipeline
    + sample["reward_model_prompt_text"])
```

maxjeblick:
Another thing I noticed is the scheduler looks a bit weird. It should be `cos(x)` with `x \in [0, pi/4]` for the whole training?
Screenshot 2023-06-20 at 19-28-17 H2O LLM Studio

pascal-pfeiffer:
Thanks a ton for your great review and all the testing. 
Great catches!

I addressed your comments and followed most of your suggestions. 
I would also like to refactor the backward pass into a reusable method, but I would do it later in subsequent issues.

maxjeblick (approved):
LGTM, thanks again for the great work!

PR #154, opened 2023-06-12
Default settings adjustment

psinger:
Default settings adjustment


psinger:
Fair point, it might be too large for smaller cards.

Without chaining it does not change much, with old settings it is still close to 30GB.

But I can revert the max length.

pascal-pfeiffer (commented):
Thank you! I agree with most changes.
This OOMs on a 24GB machine, I propose to lower the tokens again to the original values.

PR #155, opened 2023-06-12
fix for additional pretraining and more detailed FAQ

pascal-pfeiffer:
Fix for continued pretraining (failed at the prompt padding).

More detailed explanation in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,12 +321,15 @@ The majority of the LLM backbones are trained on a very similar corpus of data. 
 
 To train a chatbot style model, you need to convert your data into a question and answer format.
 
-> ❓ I encounter GPU out-of-memory issues. What can I change to be able to train large models? 
+If you really want to continue pretraining on your own data without teaching a question answering style, prepare a dataset with all your data in a single column Dataframe. Make sure that the length of the text in each row is not too long. In the experiment setup, remove all additional tokens (e.g. `<|prompt|>`, `<|answer|>`, for `Text Prompt Start` and `Text Answer Start` respectively) and disable `Add Eos Token To Prompt` and `Add Eos Token To Answer`. Deselect everything in the `Prompt Column`.
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -321,12 +321,15 @@ The majority of the LLM backbones are trained on a very similar corpus of data.
 
 To train a chatbot style model, you need to convert your data into a question and answer format.
 
-> ❓ I encounter GPU out-of-memory issues. What can I change to be able to train large models? 
+If you really want to continue pretraining on your own data without teaching a question answering style, prepare a dataset with all your data in a single column Dataframe. Make sure that the length of the text in each row is not too long. In the experiment setup, remove all additional tokens (e.g. `<|prompt|>`, `<|answer|>`, for `Text Prompt Start` and `Text Answer Start` respectively) and disable `Add Eos Token To Prompt` and `Add Eos Token To Answer`. Deselect everything in the `Prompt Column`.
```
psinger, line 324: Does this work well with LORA or does it need non-LORA?

psinger: Also shall we outsource this to maybe WIKI?

psinger: Would an empty string prompt column also work?

psinger (approved):

README.md
pascal-pfeiffer: > Also shall we outsource this to maybe WIKI?

I think it will be great in the docs (i see we have a FAQ section in the documentation PR)

pascal-pfeiffer: > Does this work well with LORA or does it need non-LORA?

It does work with LORA. I am not very sure about the quality, as this is not well tested and we have no benchmark. Will be happy to take suggestions w.r.t. automated tests/benchmarks on this.

tmm1:
Note that there's a similar FAQ entry in the docs which is now out of sync with the README.

https://github.com/h2oai/h2o-llmstudio/blob/dcbada114624435f3e38f7b97b277d1981bdc5d8/documentation/docs/faqs.md#L19-L21

PR #159, opened 2023-06-14
switch back to default chat behaviour

pascal-pfeiffer:
With the new nightly build of wave, we can switch back to the intended usage for USER and BOT in the chat window

image


maxjeblick (approved):
LGTM

PR #166, opened 2023-06-19
Add falcon peft target modules

maxjeblick:
This PR adds default lora target layers for falcon models.
I have excluded MPT models, as they seem to require additional code changes.


psinger:
@maxjeblick In best case I would envision it that we pre-fill the lora target modules field in settings automatically after switching a backbone. Do you think that's possible?

maxjeblick:
That's a good suggestion, and certainly improves UX!

Not sure what's the best way to add this, switching backbones would need to trigger an update of `ConfigNLPCausalLMTraining` parameters. Maybe add an `.update` method to `ConfigProblemBase` that can changes parameters based on current cfg and call that method before `get_ui_elements`. WDYT? It would be a bit similar to nesting, but with the ability to connect different configuration classes.

RLHF may also benefit from such an update function to hide/change certain parameters, cc. @pascal-pfeiffer

psinger:
Yes, I believe an additional functionality is needed. Would be great if you give it a shot and as you said it might be useful for other stuff.

maxjeblick:
@psinger 
The potential UI implementation mentioned above will probably take some time and effort to work correctly.
If it's ok with you, I'd merge this PR as is (or without the automatic updates, at least) to fix failing falcon experiments and work on potential UI/UX improvements afterwards.

llm_studio/src/models/text_causal_language_modeling_model.py
```diff
@@ -122,6 +110,39 @@ def __init__(self, cfg: Any):
         if self.cfg.prediction.metric == "Perplexity":
             self.perplexity = Perplexity(self.cfg, reduce=False)
 
+    def prepare_lora(self):
+        target_modules = (
+            self.cfg.training.lora_target_modules.split(",")
```
pascal-pfeiffer, line 115: Can we add a `strip` here for extra spaces in the list?

maxjeblick: Good idea, the code now applies the strip method to both `self.cfg.training.lora_target_modules` as well as each target module.

psinger (approved):

PR #178, opened 2023-06-21
fix style command

maxjeblick:
Fixes a `make style` issue I overlooked when reviewing [pp/typing](https://github.com/h2oai/h2o-llmstudio/pull/173) PR.

`$(PIPENV) run flake8` will never fail upon errors, but log them. Hence, the  command
`-$(PIPENV) run flake8 | tee -a reports/style.log || echo "flake8 failed" >> reports/style_errors.log`
 does not work as expected.


pascal-pfeiffer (approved):
right, thank you :) 
lgtm

PR #167, opened 2023-06-19
Use huggingface auth token for downloading models

maxjeblick:
Login to HF hub (using the token from the settings) when downloading weights, config and tokenizer.
Works successfully for a private model that fails on current main branch.

Fixes #143

PR #181, opened 2023-06-22
Fix auth token for falcon models

maxjeblick:
Hotfix for PR #167.

Use auth token argument is incompatible with falcon models, as their model/config/tokenizer class are missing this argument in their init. Thus, falcon models are currently failing on main.


psinger (approved):
lgtm

PR #182, opened 2023-06-22
Dtype & HF Push Changes

psinger:
This PR:

- Removes the casting to float32 in case of LORA + float16
- Adds a new mode for pushing to HF which first loads the model on CPU and then shards it across GPUs before pushing - it is probably redundant as it should do the same as CPU, but does not hurt to add it. It might be also the case that the original device is saved as a flag in the weights, which can cause downstream issues and which can be resolved via this method. In best case we would directly load the weights sharded on GPU to never be on CPU, but I did not manage to do this with LORA merging etc.


app_utils/sections/experiment.py
```diff
@@ -1900,20 +1905,33 @@ def load_cfg_model_tokenizer(experiment_path, merge=False, device="cuda:0"):
...
-            logger.info("Merging LORA layers with base model.")
-            model.backbone = model.backbone.merge_and_unload()
+    if merge and cfg.training.lora:
+        # merges the LoRa layers into the base model.
+        # This is needed if one wants to use the base model as a standalone model.
+        logger.info("Merging LORA layers with base model.")
+        model.backbone = model.backbone.merge_and_unload()
+
+    if device == "cpu_shard":
+        from accelerate import dispatch_model
```
maxjeblick, line 1922: Can we import thid on top level?

maxjeblick: Commonly, it is top level imports. I would restrict local imports to specific cases where it solves some import issue (delayed import for slow packages/side effects when importing/avoiding circular imports/ ...).

maxjeblick (approved):
LGTM, thanks!
The `UserWarning: About to update multiple times the same file in the same commit:`, seems to be an current issue with transformers not related to this PR.

app_utils/sections/experiment.py
```diff
@@ -1710,7 +1712,10 @@ async def experiment_push_to_huggingface_dialog(q: Q, error: str = ""):
                 choices=[ui.choice(str(d), str(d)) for d in devices],
                 tooltip=(
                     "The local device to prepare the model before pushing it to HF. "
-                    "CPU can be significantly slower."
+                    "CPU will never load the weights to the GPU, which can be useful "
+                    "for large models, but will be significantly slower. "
+                    "Cpu_shard will first load on CPU and then shard on all GPUs "
```
maxjeblick, line 1717: Should we name it  `gpu_shard`?

psinger: I would keep current name as it first loads on cpu and does not shard directly.

PR #192, opened 2023-06-27
Refactoring

maxjeblick:
Some light refactoring/cleanup.

- clean `.generate` method; model always gets `"prompt_input_ids"`
- move `.generate` method out of forward call in favor of calling it directly during evaluation
- raise an explicit error if config contains the words api or secret to prevent unexpected behavior when copying the config.
- Remove `num_updates` and `next(tr_iter)` in favor of enumeration loop. Refactored implementation does not 100% reproduce the original logic, as `num_updates` is tracked globally across all epochs, while `itr` is reset to 0 after each epoch. FAPP, this should not make a difference as it only is used in mod operations to check gradient accumulation updates.


pascal-pfeiffer (approved):
lgtm!

I think, you can also remove `kwargs=None` from the input vars from the `create_nlp_backbone` method, as this is never passed. 

this would further simplify
```python
if kwargs is None:
    kwargs = {}
```
to 
```python
kwargs = {}
```

PR #193, opened 2023-06-28
Def loss in dictionary

psinger:
Specifying a default loss, this is for migration reasons after we rename loss functions.


maxjeblick (approved):
LGTM, thanks!

PR #194, opened 2023-06-28
check available disk space

### LINKED ISSUE #28: [CODE IMPROVEMENT] Check for available space

We can check for available space for:

- Starting an experiment
- Pushing model to HF

### Motivation

Out of space will make an experiment fail as weights need to be saved at the end of the run; same holds for pushing to HF.

haqishen:
close #28


llm_studio/src/utils/modeling_utils.py
```diff
@@ -35,6 +36,30 @@ def unwrap_model(model: torch.nn.Module):
     return model
 
 
+def check_disk_space(model: torch.nn.Module, path: str):
```
psinger, line 39: @haqishen is this costing in terms of runtime?

did you benchmark it also on larger backbones?

haqishen: it takes almost no time for this function, like `0:00:00.007159` or so for 7b model.

psinger:
@haqishen please make a branch directly in this repo in the future, easier to review

PR #195, opened 2023-06-28
GPT metric endpoint

psinger:
Given that OpenAI regularly changes api endpoints, this PR adds an additional setting `metric_gpt_model` to specify the endpoint to be more flexible. Also fixing the default to the old gpt 3.5 model.


pascal-pfeiffer (approved):
lgtm

maybe, for downward compatibility, add default values when starting from an old experiment?

Issue #183, opened 2023-06-23
[CODE IMPROVEMENT] Check default RLHF parameters


maxjeblick:
### 🔧 Proposed code refactoring

Check if our default hyperparameters (e.g. [kl_target](https://github.com/h2oai/h2o-llmstudio/blob/main/llm_studio/python_configs/text_causal_language_modeling_config.py#L154)) are correct, see: https://github.com/lvwerra/trl/commit/b56e8b327733baa81c3ef0d6508f08e1b3e33939  and https://github.com/lvwerra/trl/issues/462

Also, RLHF training is quite unstable w.r.t. parameter choices, see e.g. issues in [trl](https://github.com/lvwerra/trl/issues). Try to find good defaults that work for one (or more) of our finetuned models.

pascal-pfeiffer:
`target_kl` is unused currently. No early stopping based on this parameter. 
Larger minibatches as default sounds good, I got the same impression w.r.t. stability there. We need another logic then to make it work on larger models without causing OOMs. The rollout is currently done in a single batched forward pass.

maxjeblick:
> target_kl is unused currently. No early stopping based on this parameter.

It is used in `AdaptiveKLController`? (as `kl_target`).

pascal-pfeiffer:
yes, these are two different params. Just wanted to make sure we talk about the same one. 
One is about early stopping (which you also linked) and the other one is the target for the controller (which is the interesting one for us).

PR #197, opened 2023-06-28
rlhf batches

pascal-pfeiffer:
adds the option to set the batch size for the PPO steps. 

partly addresses https://github.com/h2oai/h2o-llmstudio/issues/183 which suggests that larger batch sizes are more stable when fine tuning with RLHF (I can confirm)

Currently, we are limited by memory w.r.t. high batch sizes. 

also fixes a wrong value check.

We probably want to further detangle the Rollout from the PPO steps, so that we can go to higher ppo batch sizes and grad accumulations.


maxjeblick (approved):
LGTM, thanks!

PR #207, opened 2023-06-29
Diverse adjustments

psinger:
This PR makes several adjustments:

- Adds new default backbones
- Increases max token length for 8k models
- Removes redundant code for token ids
- Adds plotting utility code to decode bytes specific for certain tokenizers like xgen


llm_studio/src/utils/plot_utils.py
```diff
@@ -141,7 +141,45 @@ def format_to_html(color, word, opacity):
...
+        try:
+            # Try to decode the combined bytes
+            decoded_tokens.append(combined.decode("utf-8"))
+            # If successful, clear the buffer
+            buffer = b""
+        except UnicodeDecodeError:
+            # If decoding failed, keep the current chunk in the buffer
+            # and attempt to combine it with the next chunk
+            buffer = chunk
+        else:
```
maxjeblick, line 167: `else: ` should not be needed, as it is done in the try block.

maxjeblick (approved):
LGTM. I've left a small comment, nothing crucial.

PR #212, opened 2023-07-01
export cache dir to save downloaded backbone.

osiire:
Hi mainteners,
thank you for sharing such excellent project.

recently I have started to try H2O LLM Studio and noticed the downloaded backbone models was created in docker. so I have to download the files from Huggingface every time I wake up the docker image. that was inconvenient for me and maybe other docker lovers.

I checked the cache directory in the docker process and found it was `/home/llmstudio/.cache`. what do you think if we mount the cache directory from host?

thank you.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,3 +2,4 @@
 .github
 data
 output
+cache
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,4 +2,3 @@
 .github
 data
 output
-cache
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,6 @@ subs/
 /datasets/
 .idea/
 .local/
-cache/
 
 output
 
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ subs/
 /datasets/
 .idea/
 .local/
+cache/
 
 output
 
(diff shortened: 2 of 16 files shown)


pascal-pfeiffer (changes requested):
This is a great idea. It makes sense to have this persistent. I think we can also consider to add the cache to the `data` folder by modifying the huggingface ENV var `HUGGINGFACE_HUB_CACHE`

I think doing that in the dockerfile is the cleanest solution.

pascal-pfeiffer:
Could
```
-v ~/.cache:/home/llmstudio/.cache \
```
at Docker startup already solve what you are looking for?

osiire:
yes, the argument for docker command might be enough for many people.

the main purpose of using docker for me is keeping separate the multiple project environments from the host environment.
so I'd not mount the `/home/llmstudio/.cache` to `~/.cache`  directly, but I could change the mapping as I like if the mount point is explicitly specified.

may I modify my PR to the simple `-v ~/.cache:/home/llmstudio/.cache \` ?

pascal-pfeiffer:
Yes, I think it would help and make it clear what should be modified when one wants to change the behaviour/location of cache.

osiire:
updated.

pascal-pfeiffer (approved):
thank you, lgtm!

osiire:
I'm sorry I closed the PR before merge by mistake. reopned.

PR #217, opened 2023-07-02
Customize data_folder & output_folder with environment variables (#216)

### LINKED ISSUE #216: [FEATURE] Customize data_folder & output_folder with environment variables
### 🚀 Feature

Customize https://github.com/h2oai/h2o-llmstudio/blob/main/app_utils/config.py#L48-L49 with environment variables.

### Motivation

Allow customizing `data_folder` and `output_folder`  without needing to edit `app_utils/config.py` code.

My use case is Runpod only having a single volume option so I want to override the `data_folder` and `output_folder` to point inside a single persisted volume.

Glavin001:
Closes #216

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,6 +344,7 @@ H2O LLM Studio stores its data in two folders located in the root directory. The
 
 It is possible to change the default folders `data` and `output` in [app_utils/config.py](app_utils/config.py) 
 (change `data_folder` and `output_folder`).
+Or set `DATA_FOLDER` and `OUTPUT_FOLDER` environment variables.
 
 > ❓ How can I update H2O LLM Studio?
 
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ The interactive chat will also work with model that were finetuned using the UI.
 
 H2O LLM studio expects a csv file with at least two columns, one being the instruct column, the other being the answer that the model should generate. You can also provide an extra validation dataframe using the same format or use an automatic train/validation split to evaluate the model performance.
 
-During an experiment you can adapt the data representation with the following settings 
+During an experiment you can adapt the data representation with the following settings
 
 - **Prompt Column:** The column in the dataset containing the user prompt.
 - **Answer Column:** The column in the dataset containing the expected output.
@@ -339,14 +339,13 @@ LORA should be enabled. Besides that there are the usual parameters such as batc
 > ❓ Where does H2O LLM Studio store its data?
 
 H2O LLM Studio stores its data in two folders located in the root directory. The folders are named `data` and `output`. Here is the breakdown of the data storage structure:
+
 - `data/dbs`: This folder contains the user database used within the app.
 - `data/user`: This folder is where uploaded datasets from the user are stored.
 - `output/user`: All experiments conducted in H2O LLM Studio are stored in this folder. For each experiment, a separate folder is created within the `output/user` directory, which contains all the relevant data associated with that particular experiment.
-- `output/download`: Utility folder that is used to store data the user downloads within the app. 
+- `output/download`: Utility folder that is used to store data the user downloads within the app.
 
-It is possible to change the default folders `data` and `output` in [app_utils/config.py](app_utils/config.py) 
-(change `data_folder` and `output_folder`).
-Or set `DATA_FOLDER` and `OUTPUT_FOLDER` environment variables.
+It is possible to change the default folders `data` and `output` in [app_utils/config.py](app_utils/config.py) (change `data_folder` and `output_folder`). Or set `LLM_STUDIO_DATA_FOLDER` and `LLM_STUDIO_OUTPUT_FOLDER` environment variables.
 
 > ❓ How can I update H2O LLM Studio?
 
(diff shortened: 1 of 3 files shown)


pascal-pfeiffer (changes requested):
Great idea!

Let's make the ENV variables a bit less generic.

I also think that it may cause issues if both point to the same ressource w.r.t. file names. For example if you name an experiment `user`

app_utils/config.py
```diff
@@ -45,8 +45,8 @@ def get_size(x):
     "allowed_file_extensions": os.getenv(
         "ALLOWED_FILE_EXTENSIONS", ".zip,.csv,.pq"
     ).split(","),
-    "data_folder": "data/",
-    "output_folder": "output/",
+    "data_folder": f"{os.getenv('DATA_FOLDER', 'data/')}",
+    "output_folder": f"{os.getenv('OUTPUT_FOLDER', 'output/')}",
```
Glavin001, line 49: Definitely agree @pascal-pfeiffer . Any suggestions?

For example:

```suggestion
    "output_folder": f"{os.getenv('LLM_STUDIO_OUTPUT_FOLDER', 'output/')}",
```

pascal-pfeiffer:
```python
"output_folder": str(os.getenv("LLM_STUDIO_OUTPUT_FOLDER", os. getcwd())) + "/output/"
```

Name is good. What about something along those lines? It should allow to use a single volume for both.

I think the Dockerfile will also need to be touched, though.

Glavin001: This would enforce the end directory is always named exactly “output”. Works for the Docker image (/workspace/output) and my use case — single volume with 2 folders within — although would mean users are unable to change the name from “output” if they needed to. 

My preference is to keep it simple and flexible by letting the entire value be able to be overwritten. Although this works for me 👍 

+1 for defaulting to cwd. 

How about: 

```suggestion
"output_folder": str(os.getenv("LLM_STUDIO_OUTPUT_FOLDER", f”{os.getcwd()}/output/"))
```

?

pascal-pfeiffer: I also like to have it as flexible as possible. The issue is, that people may use the same ENV variable for both folders and that may lead to very unexpected results (e.g. experiment being named `user`). I think my proposal would only create one more folder nesting and as you said, enforce end directories to be named `output` and `data`.

Ultimately, I would probably adapt the basic Docker approach to use a single mounted volume.
So, we can replace these two lines with a single volume that is attached to the Docker image

```
    -v `pwd`/data:/workspace/data \
    -v `pwd`/output:/workspace/output \
```

with

```
    -v `pwd`:/workspace\
```

while also allowing to pass the ENV var you added to the Docker image and customizing the folders individually for very specific use cases where the two folders need to sit on different volumes. 


There is also this related PR (https://github.com/h2oai/h2o-llmstudio/pull/212) suggesting to use the hosts `.cache` folder.

Glavin001: > people may use the same ENV variable for both folders and that may lead to very unexpected results

Ah I see, eliminate the possibility of human-error, don't want to end up with:

```yaml
output_folder: /workspace
data_folder: /workspace
```

by mistake.

Glavin001: > So, we can replace these two lines with a single volume that is attached to the Docker image

Given https://github.com/h2oai/h2o-llmstudio/blob/main/Dockerfile#L28-L37 I think the `/workspace` directory is not clean / contains **_more_** than only `data/` and `output/`, which would be deleted by the volume:

```bash
    -v `pwd`:/workspace\
```

- `WORKDIR` and therefore `pwd` is `/workspace`: https://github.com/h2oai/h2o-llmstudio/blob/main/Dockerfile#L27C1-L27C1

Therefore,

```bash
    -v `pwd`/data:/workspace/data \
    -v `pwd`/output:/workspace/output \
```

is still appropriate, ✅  no change required.

> Ultimately, I would probably adapt the basic Docker approach to use a single mounted volume.

This could be changed in another PR? I'd want to avoid breaking changes with this PR and also see if there are other improvements (e.g. #212 ) to be considered before improving the Docker approach.

Glavin001: - Pushed change based on [your proposal above](https://github.com/h2oai/h2o-llmstudio/pull/217#discussion_r1251499947) which is less human error prone
- Updated to latest `main`

Glavin001: - Pushed fix to Black formatting errors

Glavin001:
CI passing. Ready for another review and hopefully merge 😃  Thanks!

app_utils/config.py
Glavin001: Any other changes requested? 😃

pascal-pfeiffer:
thank you, lgtm!
Now we can think about changing the Dockerfile in a follow-up PR

pascal-pfeiffer (approved):

README.md
```diff
@@ -346,6 +346,7 @@ H2O LLM Studio stores its data in two folders located in the root directory. The
 
 It is possible to change the default folders `data` and `output` in [app_utils/config.py](app_utils/config.py) 
 (change `data_folder` and `output_folder`).
+Or set `DATA_FOLDER` and `OUTPUT_FOLDER` environment variables.
```
pascal-pfeiffer, line 349: LLM_STUDIO_DATA_FOLDER and LLM_STUDIO_OUTPUT_FOLDER

documentation/docs/faqs.md
```diff
@@ -50,6 +50,7 @@ H2O LLM Studio stores its data in two folders located in the root directory. The
 
 It is possible to change the default folders `data` and `output` in [app_utils/config.py](https://github.com/h2oai/h2o-llmstudio/blob/main/app_utils/config.py) 
 (change `data_folder` and `output_folder`).
+Or set `DATA_FOLDER` and `OUTPUT_FOLDER` environment variables.
```
pascal-pfeiffer, line 53: LLM_STUDIO_DATA_FOLDER and LLM_STUDIO_OUTPUT_FOLDER

pascal-pfeiffer (changes requested):
two minor docs changes please before merging

pascal-pfeiffer (approved):

Glavin001:
Ah forgot about the docs. Thank you for fixing up and merging! 🎉

PR #221, opened 2023-07-03
add: safe serialization while pushing to hf

### LINKED ISSUE #187: [FEATURE] Push Safetensor to HF
Explore to always push Safetensor to HF.

shivance:
Closes #187

### DIFF
--- a/app_utils/sections/experiment.py
+++ b/app_utils/sections/experiment.py
@@ -1777,6 +1777,7 @@ async def experiment_push_to_huggingface_dialog(q: Q, error: str = ""):
         tokenizer.push_to_hub(
             repo_id=repo_id,
             private=True,
+            safe_serialization=True
         )
 
         # push model card to hub
@@ -1785,6 +1786,7 @@ async def experiment_push_to_huggingface_dialog(q: Q, error: str = ""):
             repo_id=repo_id,
             repo_type="model",
             commit_message="Upload model card",
+            safe_serialization=True
         )
 
         # push config to hub
@@ -1808,6 +1810,7 @@ async def experiment_push_to_huggingface_dialog(q: Q, error: str = ""):
             repo_id=repo_id,
             private=True,
             commit_message="Upload model",
+            safe_serialization=True
         )
 
         # push pipeline to hub


psinger:
Please run `make format` and update the branch.

shivance:
@psinger all checks pass !

app_utils/sections/experiment.py
```diff
@@ -1774,17 +1774,15 @@ async def experiment_push_to_huggingface_dialog(q: Q, error: str = ""):
         repo_id = f"{user_id}/{exp_name}"
 
         # push tokenizer to hub
-        tokenizer.push_to_hub(
-            repo_id=repo_id,
-            private=True,
-        )
+        tokenizer.push_to_hub(repo_id=repo_id, private=True, safe_serialization=True)
```
pascal-pfeiffer, line 1777: I think it will just be ignored here, so let's probably remove to avoid confusion,

pascal-pfeiffer (changes requested):
Thank you @shivance, great find! 
I think this is not needed or even buggy when being used with tokenizer and model card.

For the model weights, can we make it a checkbox in the upload dialog (probably True by default is OK)?

shivance:
Hi @pascal-pfeiffer !
Removed those args from `tokenizer` and `model_card`.

app_utils/sections/experiment.py
```diff
@@ -1734,6 +1734,7 @@ async def experiment_push_to_huggingface_dialog(q: Q, error: str = ""):
                 required=True,
                 tooltip="HF API key, needs write access.",
             ),
+
```
shivance, line 1737: @pascal-pfeiffer I should add a ui.checkbox here right? And how do we get a placeholder based on it's value which in turn has to be passed to 
```
        model.backbone.push_to_hub(
            repo_id=repo_id,
            private=True,
            commit_message="Upload model",
            safe_serialization=<placeholder>,
        )
```

PS: Found that to add checkbox in h2o-wave we use ui.checkbox
https://github.com/h2oai/wave/blob/7cf6b29c342771cf0e2a7e81ac93c121715338a1/website/widgets/form/checkbox.md?plain=1#L31

shivance:
Also is there a way to run format without make, it's throwing some errors for me now, I'm using a custom conda environment.

psinger:
@shivance custom conda environment is not recommended, specifically for development.
Best is to run `make setup` and then use that environment.

app_utils/sections/experiment.py
pascal-pfeiffer: yes, you can use `ui.checkbox`, give it a name and use q.client.NAME_OF_CHECKBOX value for your `<placeholder>`. 

`copy_expando(q.args, q.client)` in app.py makes sure that q.client values are always the latest submitted values from q.args

pascal-pfeiffer:
> Also is there a way to run format without make, it's throwing some errors for me now, I'm using a custom conda environment.

the make command basically runs

```sh
python run isort .
python run black .
```

so, you can also run that you own, but as @psinger said, using the pipenv is highly recommended over a custom conda environment. For development, you will need to run `make setup-dev` to get the linter etc.

shivance:
Hi @psinger & @pascal-pfeiffer Could you review?
Instead of checkbox how about we use a `toggle` ?

pascal-pfeiffer:
> Hi @psinger & @pascal-pfeiffer Could you review? Instead of checkbox how about we use a `toggle` ?

Thank you for the changes. Toggle is a great idea, probably better UX.

shivance:
CI is ✅

pascal-pfeiffer (changes requested):
Thank you @shivance !

Almost good to be merged. Two comments in the code.

app_utils/sections/experiment.py
```diff
@@ -1535,6 +1535,12 @@ async def experiment_push_to_huggingface_dialog(q: Q, error: str = ""):
                 required=True,
                 tooltip="HF API key, needs write access.",
             ),
+            ui.toggle(
+                name="default_safe_serialization",
+                label="Use Huggingface safetensors for safe serialization",
+                value=q.client["default_safe_serialization"],
+                trigger=True,
```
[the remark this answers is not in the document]
shivance: So I shall remove the trigger key value pair from here right?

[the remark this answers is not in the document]
pascal-pfeiffer: yes, I think just removing the trigger would already solve it, as `q.client["default_safe_serialization"]` will be updated when the user clicks "Export"

pascal-pfeiffer (approved):
Thanks again for the great find, lgtm!

PR #222, opened 2023-07-03
Checkpoint changes

psinger:
This PR changes the following:

- Removes nesting for best checkpoint saving
- Save last checkpoint after each validation, this can be useful for long runs with many sub-validations where it is risky to only save the checkpoint in the very end. extra runtime should be minimal


maxjeblick (approved):
lgtm

Issue #213, opened 2023-07-01
[BUG] Pushing int8 model to Huggingface has error: AttributeError: 'str' object has no attribute 'dtype'


Glavin001:
## How to Reproduce

1. Train
2. Push to Huggingface
3. Error 😢 

With:
- Docker image: <TODO>
- Runpod A100 80GB

Config:

```yaml
architecture:
    backbone_dtype: int8
    force_embedding_gradients: false
    gradient_checkpointing: true
    intermediate_dropout: 0.0
    pretrained: true
    pretrained_weights: ''
augmentation:
    random_parent_probability: 0.0
    skip_parent_probability: 0.0
    token_mask_probability: 0.0
dataset:
    add_eos_token_to_answer: true
    add_eos_token_to_prompt: true
    answer_column: output
    chatbot_author: H2O.ai
    chatbot_name: h2oGPT
    data_sample: 1.0
    data_sample_choice:
    - Train
    - Validation
    limit_chained_samples: false
    mask_prompt_labels: true
    parent_id_column: None
    personalize: false
    prompt_column:
    - instruction
    text_answer_separator: <|answer|>
    text_prompt_start: <|prompt|>
    train_dataframe: data/user/startup-interviews/json-train.pq
    validation_dataframe: None
    validation_size: 0.01
    validation_strategy: automatic
environment:
    compile_model: false
    find_unused_parameters: false
    gpus:
    - '0'
    huggingface_branch: main
    mixed_precision: true
    number_of_workers: 8
    seed: -1
    trust_remote_code: true
    use_fsdp: false
experiment_name: startup-interviews
llm_backbone: huggyllama/llama-13b
logging:
    logger: Neptune
    neptune_project: glavin001/startup-interviews
    number_of_texts: 10
output_directory: output/user/startup-interviews/
prediction:
    batch_size_inference: 0
    do_sample: false
    max_length_inference: 256
    metric: BLEU
    min_length_inference: 2
    num_beams: 1
    num_history: 2
    repetition_penalty: 1.2
    stop_tokens: ''
    temperature: 0.3
    top_k: 0
    top_p: 1.0
tokenizer:
    add_prefix_space: false
    add_prompt_answer_tokens: false
    max_length: 2048
    max_length_answer: 1808
    max_length_prompt: 192
    padding_quantile: 1.0
    use_fast: false
training:
    adaptive_kl_control: true
    advantages_gamma: 0.99
    advantages_lambda: 0.95
[...]

PR #226, opened 2023-07-03
Bitsandbytes fixes

psinger:
Closes https://github.com/h2oai/h2o-llmstudio/issues/213

Recent BNB update introduced some extra keys in the state dicts without any tensor `weight_format`
need to filter them also when loading with float16

Also noticed another issue that int8 weights could also not be loaded anylonger because BNB hardcodes the same devices as is present in the weights file which for us always was cpu. So now I am taking the device specified in cfg to directly load the weights to the device.

### DIFF
--- a/.github/ISSUE_TEMPLATE/doc-request.md
+++ b/.github/ISSUE_TEMPLATE/doc-request.md
@@ -1,0 +1,15 @@
+---
+name: "\U0001F41B Documentation request"
+about: Create a doc request
+title: "[DOC]"
+labels: type/doc
+assignees: ''
+---
+
+### 📃 Documentation
+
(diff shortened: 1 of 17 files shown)


maxjeblick (changes requested):
Thanks a lot for the fixes!

Regarding the loading of the model weights, I observed increased peak GPU memory consumption (17404MiB on main vs. 30172MiB on this branch) when loading a model in the chat window.

psinger:
Hmh thats strange, will need to investigate, thx.

psinger:
Loading it on cpu seems better... otherwise it can need extra mem on GPU.
But then need to figure out the BNB issue... hard

maxjeblick (approved):
lgtm, thanks!

PR #230, opened 2023-07-03
Fix chat bug

### LINKED ISSUE #215: [BUG] Chat tab broken and loading forever.
### 🐛 Bug

Chat tab broken and loading forever.

image

### To Reproduce

1. Train model
2. Open `Chat` tab
3. It successfully loads the `Chat Settings` at the bottom
4. Never finishes loading the upper chat conversation UI. Says `Loading...` forever.
5. Submit a chat message and wait.
6. Look at the `_s/` WebSocket `Messages` and see ✅ your message was sent and ✅ response was generated and received. See screenshot ☝️ .

maxjeblick:
Fixes a chat window bug that occurs when the user enters a question and presses enter before the model was loaded.
In that case, `cfg_prediction = parse_ui_elements...` is not handled correctly, as the rendering of the chat settings seems to be skipped.

This PR replaces the chat window with a progress window such that the chat window becomes only available after the model has been loaded.

Fixes #215

### DIFF
--- a/app_utils/sections/chat.py
+++ b/app_utils/sections/chat.py
@@ -79,7 +79,7 @@ async def chat_tab(q: Q, load_model=True):
 
     if load_model:
         q.page["experiment/display/chat"] = ui.form_card(
-            box="second",
+            box="first",
             items=[ui.progress(label="Loading the model...")],
         )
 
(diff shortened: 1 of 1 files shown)


psinger (approved):
LGTM!

Issue #227, opened 2023-07-03
[BUG] CUDA Error when using 'Add Prompt Answer Tokens' 


luizlf:
### 🐛 Bug

Got the error below for an experiment with backbone `tiiuae/falcon-7b-instruct` and option 'Add Prompt Answer Tokens' set to True. The error happens with 'Force Embedding Gradients' set to either True or False.

`2023-07-03 11:40:25,696 - INFO: Number of observations in train dataset: 44778
2023-07-03 11:40:25,696 - INFO: Number of observations in validation dataset: 453
2023-07-03 11:40:26,399 - INFO: dtype: torch.float16
2023-07-03 11:40:43,594 - INFO: Training Epoch: 1 / 1
2023-07-03 11:40:43,595 - INFO: train loss: 0%| | 0/14926 [00:00
2023-07-03 11:40:44,483 - INFO: Stop token ids: [tensor([65025]), tensor([65024])]
2023-07-03 11:40:44,693 - ERROR: Exception occurred during H2O LLM Studio run:
Traceback (most recent call last):
File "/home/fsuser/projects/h2o-llmstudio/train_wave.py", line 106, in
run(cfg=cfg)
File "/home/fsuser/projects/h2o-llmstudio/train.py", line 570, in run
val_data, val_loss, val_metric, last_batch = run_train(
File "/home/fsuser/projects/h2o-llmstudio/train.py", line 271, in run_train
output_dict = model.forward(batch)
File "/home/fsuser/projects/h2o-llmstudio/llm_studio/src/models/text_causal_language_modeling_model.py", line 193, in forward
output = self.backbone(
File "/home/fsuser/.local/share/virtualenvs/h2o-llmstudio-XOsEFuK8/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1501, in _call_impl
return forward_call(*args, **kwargs)
File "/home/fsuser/.local/share/virtualenvs/h2o-llmstudio-XOsEFuK8/lib/python3.10/site-packages/peft/peft_model.py", line 678, in forward
return self.base_model(
File "/home/fsuser/.local/share/virtualenvs/h2o-llmstudio-XOsEFuK8/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1501, in _call_impl
return forward_call(*args, **kwargs)
[...]

PR #231, opened 2023-07-03
Fix tokenizer length

psinger:
Closes https://github.com/h2oai/h2o-llmstudio/issues/227

old implementation did not take new tokens into account

https://github.com/huggingface/transformers/blob/v4.30.0/src/transformers/tokenization_utils_fast.py#L143
https://github.com/huggingface/transformers/blob/v4.30.0/src/transformers/tokenization_utils_fast.py#L168


pascal-pfeiffer (approved):
Great catch, thank you @psinger 

lgtm!

PR #234, opened 2023-07-03
cast to float before converting to numpy

### LINKED ISSUE #233: [BUG] Perplexity metric fails when using BFloat16 Backbone dtype
### 🐛 Bug

When using bfloat16 for the Backbone dtype, the Perplexity validation metric cannot be evaluated, yielding the following error:

```
  File "/home/ubuntu/h2o-llmstudio/llm_studio/src/metrics/text_causal_language_modeling_metrics.py", line 149, in perplexity
    return results["perplexity"].detach().cpu().numpy()
TypeError: Got unsupported ScalarType BFloat16
```

To fix, we can convert the perplexity tensor to float32 before converting to numpy, i.e.

```
results["perplexity"] = results["perplexity"].to(torch.float32)
return results["perplexity"].detach().cpu().numpy()
```

### To Reproduce

Create an experiment using BFloat16 as the backbone dtype, and perplexity as the metric. The above error will be thrown on the first validation pass.

pascal-pfeiffer:
closes #233

### DIFF
--- a/llm_studio/src/metrics/text_causal_language_modeling_metrics.py
+++ b/llm_studio/src/metrics/text_causal_language_modeling_metrics.py
@@ -146,7 +146,7 @@ class Perplexity(nn.Module):
 
 
 def perplexity(cfg: Any, results: Dict, val_df: pd.DataFrame):
-    return results["perplexity"].detach().cpu().numpy()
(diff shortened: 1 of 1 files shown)


maxjeblick (approved):
lgtm

PR #236, opened 2023-07-03
make lora slider max values adjustable

### LINKED ISSUE #235: [FEATURE] Allow setting higher LORA rank and alpha values from the UI
### 🚀 Feature

Allow larger LORA rank and alpha values in the UI (at least 512 and 1024 respectively).

### Motivation

I've found that in some cases using a larger LORA rank and alpha than allowed in the UX (which limits rank to a max of 256 and alpha to a max of 512) can improve model performance. Specifically, I've been using a LORA rank of 512 and alpha of 1024 with good results. I tweaked `text_causal_language_modeling_config.py` locally to allow me to configure these values from the UI, but it would be nice if this worked out of the box.

pascal-pfeiffer:
closes #235

### DIFF
--- a/app_utils/config.py
+++ b/app_utils/config.py
@@ -89,6 +89,8 @@ default_cfg = {
         "set_max_epochs": 50,
         "set_max_batch_size": 256,
         "set_max_gradient_clip": 10,
+        "set_max_lora_r": 256,
+        "set_max_lora_alpha": 256,
         "gpu_used_for_chat": 1,
         "default_number_of_workers": 8,
         "default_logger": "None",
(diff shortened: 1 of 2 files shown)


psinger:
@pascal-pfeiffer maybe we can just increase the max values? do we really need a setting here?

pascal-pfeiffer:
It becomes harder to set specific low values if the max value is really large. Let's say 2048 is the max value, it's hard to set it to 4. 
I don't think it is too common to set it to very high values (right now), so I would hide it a bit.

maxjeblick (approved):
lgtm! Regarding the restructuring of max settings, we could track that in a subsequent issue/PR.

Issue #240, opened 2023-07-04
[BUG] Inconsistency between Validation answers and Chat answers


aerdem4:
### 🐛 Bug

For the same question, validation set has different answers compared to asking it on Chat screen. Both have no previous context. Chat screen answers look better.

I have read the code extensively. Different pipelines are used for run_eval and chat but I couldn't find what is really different between them.

### To Reproduce

I use falcon-7b and validation set examples have tendency to use a lot of hashtags at the end of answers. Same doesn't happen during chat.

maxjeblick:
Thanks for opening this issue, it's a really good catch!

I digged a bit into the code, here are some observations:

1. Previous config items in the chat tab are preferred over the model configuration. Thus, switching between models or changing a parameter and then reloading the chat tab may result in generation configs that is not equivalent to the generation config used during training.
2.  Surprisingly, the falcon model seems to be quite sensitive to (left) padding. I tried the prompt `Hi! I would like to know if you can roleplay a human for me.` (from OASST dataset) which gives the following answers (temperature is set to 0 to avoid randomness):

    -    **Chat UI:** `Hello, I am a human named h2oGPT. How may I assist you today?`
    -    **Validation Prediction:** `Hello, I am a human. How can I help you?`

    When I add manual left padding to the [code for chat generation](https://github.com/h2oai/h2o-llmstudio/blob/main/app_utils/sections/chat.py#L229): ```full_prompt = tokenizer.pad_token * 20 + full_prompt```, the chat window outputs the same prediction as during evaluation. This could be an issue with falcon's implementation of attention, seems as if attention mask is missing when computing `F.scaled_dot_product_attention`. I will further look into this.

aerdem4:
Thanks for looking into it!

Mine doesn't match even if I set batch size to 1. But I try this after finetuning. Did you finetune before trying?

Attention mask is used unless alibi is None. (What is alibi?) Its tokenizer doesnt have a mask token, and last token is ' victoria'. So the code below puts a lot of victorias probably?

```
    if tokenizer.unk_token_id is not None:
        cfg._tokenizer_mask_token_id = tokenizer.unk_token_id
    elif tokenizer.mask_token_id is not None:
        cfg._tokenizer_mask_token_id = tokenizer.mask_token_id
    elif tokenizer.pad_token_id is not None:
        cfg._tokenizer_mask_token_id = tokenizer.pad_token_id
    else:
        # setting the mask token id to the last token in the vocabulary
[...]

PR #248, opened 2023-07-05
Generation config fix

psinger:
Discussion here:
https://github.com/h2oai/h2o-llmstudio/issues/240

### DIFF
--- a/llm_studio/src/datasets/text_utils.py
+++ b/llm_studio/src/datasets/text_utils.py
@@ -58,13 +58,11 @@ def get_tokenizer(cfg: Any):
     if tokenizer.eos_token == "":
         tokenizer.add_special_tokens({"eos_token": "</s>"})
         tokenizer.eos_token = "</s>"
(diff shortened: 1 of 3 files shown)


psinger:
Agree on the first suggestion, will change.

But we have the `load_cfg_model_tokenizer` helper function that should already make sure this is handled. Creating the tokenizer twice sounds inefficient to me.

maxjeblick:
> But we have the load_cfg_model_tokenizer helper function that should already make sure this is handled. Creating the tokenizer twice sounds inefficient to me.

It does work with the existing codebase. Yet, it creates some additional entanglement across the codebase, e.g. I could not initialize `Model` class in isolation. As calling `get_tokenizer` introduces little computational overhead, I'd prefer this method.

psinger:
ok I decided to take it from model config and make it transparent, wdyt? too much?

maxjeblick (approved):
Lgtm!

Issue #223, opened 2023-07-03
[FEATURE] Separate functionality for system prompt training


psinger:
### 🚀 Feature

Add a separate setting to specify system prompt and corresponding system special token.

Such as
`<|system|>system prompt<|prompt|>prompt<|answer|>`

### Motivation

While prepending system prompt is currently supported, more flexible support might be useful.

shivance:
Hi @psinger can I contribute to this?

psinger:
@shivance thanks for volunteering, but I will take this one

please consider taking another open issue

PR #250, opened 2023-07-05
System prompt

psinger:
Closes https://github.com/h2oai/h2o-llmstudio/issues/223

Currently not working for rlhf

Tooltips will follow after final decisions are made

### DIFF
--- a/llm_studio/src/datasets/text_causal_language_modeling_ds.py
+++ b/llm_studio/src/datasets/text_causal_language_modeling_ds.py
@@ -81,7 +81,7 @@ class CustomDataset(Dataset):
                 self.df_id_to_idx = {v: k for k, v in enumerate(self.df["id"].values)}
 
                 # limit chained samples to the longest chain
-                if self.cfg.dataset.limit_chained_samples:
+                if self.cfg.dataset.limit_chained_samples and self.mode == "train":
                     unique_parent_ids = set(self.parent_ids)
                     self.indices = self.indices[
                         [id not in unique_parent_ids for id in self.df["id"].values]
--- a/llm_studio/src/datasets/text_causal_language_modeling_ds.py
+++ b/llm_studio/src/datasets/text_causal_language_modeling_ds.py
@@ -82,8 +82,9 @@ class CustomDataset(Dataset):
 
                 # limit chained samples to the longest chain
                 if self.cfg.dataset.limit_chained_samples:
+                    unique_parent_ids = set(self.parent_ids)
                     self.indices = self.indices[
-                        [id not in self.parent_ids for id in self.df["id"].values]
+                        [id not in unique_parent_ids for id in self.df["id"].values]
                     ]
 
         self.systems = None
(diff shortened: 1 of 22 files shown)


llm_studio/src/datasets/text_causal_language_modeling_ds.py
```diff
@@ -304,23 +335,32 @@ def sanity_check(cls, df: pd.DataFrame, cfg: Any, mode: str = "train"):
...
         encodings = parent_encodings + encodings
 
-        sample["reward_model_prompt_text"] = (
-            reward_model_parent_prompt_text + self.raw_prompts[idx]
-        )
+        # get first system encoding
+        system_encoding = encodings[0][0]
+        # remove system encodings from list of encodings
+        encodings = [encoding[1:] for encoding in encodings]
+        encodings[0][0] = torch.cat([system_encoding, encodings[0][0]])
```
pascal-pfeiffer, line 358: I don't fully grasp that part and why we append it already here. Should it be prepended at the end maybe?

llm_studio/src/possible_values.py
```diff
@@ -25,7 +25,7 @@ def _scan_dirs(dirname) -> List[str]:
 
 
 def _scan_files(
-    dirname, extensions: Tuple[str, str, str] = (".csv", ".pq", ".json")
+    dirname, extensions: Tuple[str, ...] = (".csv", ".pq", ".parquet", ".json")
```
pascal-pfeiffer, line 28: :+1:

pascal-pfeiffer (changes requested):
Thanks a lot for the functionality @psinger 
- I would suggest that we also add it to the validation visualizations. 
- Can we add it to the chat interface somehow?
- as you said, tooltips are missing

I have a bit of understanding issues why we append the systems prompt embedding as early. Probably a bit tricky with the chained conversations.

llm_studio/src/datasets/text_causal_language_modeling_ds.py
psinger: So currently each sample in `encodings` has the system prompt, but we only need it for the root, which is why we pick it from `encodings[0][0]` and then remove it again, so that we only have prompt/answer as before.
Then we just concat it to the very first prompt and have the typical workflow.

Then we only need to have this extra functionality at the end, to make sure it is prepended if max token length is exceeded. Otherwise we would always need to prepend it manually, which we could also do, but a bit more messy because we would need to do it after the padding, and then we would need to find the index where the padding ends etc.

Bit messy, but "simplest" solution I could think of. Im sure there are better ways.

psinger:
> Can we add it to the chat interface somehow?

We could, but do you think it would be an important feature? would need to be an extra input then

> I would suggest that we also add it to the validation visualizations.

Lets add it here: https://github.com/h2oai/h2o-llmstudio/issues/255

llm_studio/src/datasets/text_causal_language_modeling_ds.py
pascal-pfeiffer: Thank you, now I understand. Makes perfect sense.
Could you maybe add a few short comments in the code for that reasoning. Basically what you said above.

pascal-pfeiffer:
> > Can we add it to the chat interface somehow?
> 
> We could, but do you think it would be an important feature? would need to be an extra input then
> 
> > I would suggest that we also add it to the validation visualizations.
> 
> Lets add it here: #255


Agree, this can definitely be done in follow-up issues.
Else, looks good to me, besides the missing tooltips.

PR #177, opened 2023-06-21
Dataset refactoring

maxjeblick:
This PR refactors some of the dataset methods for better readability.

PR #256, opened 2023-07-06
Fix limit_chained_samples

maxjeblick:
Fixes a bug that was introduced in #177, causing an incorrect behavior of `limit_chained_samples`.


psinger (approved):
should fix it, ty

PR #265, opened 2023-07-10
Token IDs (again)

psinger:
@maxjeblick this is another attempt to change the token ids for the configs basically following your original suggestion

I just had a thought that the old approach would probably still fail if `config` and `generation_config` token ids are wrong

wdyt?


maxjeblick (approved):
lgtm, thanks! We could think about extracting a common function to overwrite/fix the configurations, but this is not crucial.

PR #269, opened 2023-07-10
end user shall use "make llmstudio" to run without wave checking for file changes

### LINKED ISSUE #267: [BUG] Wave reload triggers when uploading the same model twice to Hugging Face
### 🐛 Bug

There is a small issue that I found when trying to upload the same model twice. The wave live update will trigger due to changes in the generated pipeline code.

### To Reproduce

Uplaod same model twice to Hugging Face without deleting the pipeline python file in between.

pascal-pfeiffer:
H2O Wave monitors the app dir for code changes by default since v0.0.9 (https://wave.h2o.ai/blog/release-0.9.0#live-reload).

We create the `h2oai_pipeline.py` file in the output directory when uploading a model to Hugging Face. When doing so a second time, this triggers the reload of wave. 

There is currently no way to disable the check for certain folders, so I decided to disable the reload check completely when not developing. For that, the end user will start the app using `make llmstudio` which starts the wave server without reloading (`--no-reload`).

closes #267

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,8 +80,8 @@ wave:
 	H2O_WAVE_PRIVATE_DIR="/download/@$(WORKDIR)/output/download" \
 	$(PIPENV) run wave run app
 
-.PHONY: wave-no-reload
-wave-no-reload:
+.PHONY: llmstudio
+llmstudio:
 	H2O_WAVE_MAX_REQUEST_SIZE=25MB \
 	H2O_WAVE_NO_LOG=true \
 	H2O_WAVE_PRIVATE_DIR="/download/@$(WORKDIR)/output/download" \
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -80,8 +80,8 @@ wave:
 	H2O_WAVE_PRIVATE_DIR="/download/@$(WORKDIR)/output/download" \
 	$(PIPENV) run wave run app
 
-.PHONY: wave-no-reload
-wave-no-reload:
+.PHONY: llmstudio
```
maxjeblick, line 83: Like the new name!  Should we remove or rename `make wave`, e.g. `make llmstudio-reload` ?

pascal-pfeiffer: feels natural :) 

actually, `make wave` is already in my muscle memory and quite quick to type, so I kind of like it for development. But I would also be fine with `make llmstudio-reload` if you have a strong preference.

maxjeblick: Yeah, I'm also used to it :D Thus, free to leave as is.

maxjeblick (approved):
lgtm thanks

PR #273, opened 2023-07-11
Fix neptune version

psinger:
Neptune 1.3.2 is completely broken, no idea how they could release it in that state

I cant see any release logs detailing any required changes, also it is just a minor release


pascal-pfeiffer (approved):
good find, thank you @psinger

Issue #171, opened 2023-06-20
Improve Model Card Template


psinger:
Here are a few things to improve in the model card:

- Some models fail if input contains wrong keys, such as type ids. We should delete them after the tokenizer, or be specific of what we send to the `generate`, i.e. only send input ids and attention mask.
- Remove model validation section
- Potentially add quantization example everywhere

Glavin001:
Applicable code: https://github.com/h2oai/h2o-llmstudio/blob/25164164fd36c5e9ac8ce39275a709e649516531/model_card_template.md

- [ ] Replace example prompt (i.e. `"Why is drinking water so healthy?"`, `How are you?`) with an actual example prompt from the dataset.

psinger:
I think having a static example is a better idea here as it is hard to control and you might not want your data to leak outside.

PR #274, opened 2023-07-11
Minor model template changes

psinger:
Closes https://github.com/h2oai/h2o-llmstudio/issues/171


pascal-pfeiffer (approved):
lgtm

shall we add einops dependency for falcon models, too?

pascal-pfeiffer (approved):
thank you @psinger 

we could make it optional and only add for falcon backbone models, but I think this is good. It's a rather common library and often used with pytorch, so there may be more models depending on it in the future.

Issue #280, opened 2023-07-13
[BUG] Fix support for Runpod: Permission denied


Glavin001:
### 🐛 Bug

There are multiple issues getting the latest LLM Studio working in Runpod.

- :x: `gcr.io/vorvan/h2oai/h2o-llmstudio:nightly`
- ✅ `gcr.io/vorvan/h2oai/h2o-llmstudio@sha256:8c3c86d0e2721c35803924ee6005ccf0ebd8bc3b2568bdca5ca5d202b2074a1c` (before #163 )

#### Root cause

#163 and Runpod does not appear to have a way to add to `docker` call:

```
-u `id -u`:`id -g`
```

so we need workarounds.

### Workaround

The following template uses an older Docker image of LLM Studio and therefore works.

`gcr.io/vorvan/h2oai/h2o-llmstudio@sha256:8c3c86d0e2721c35803924ee6005ccf0ebd8bc3b2568bdca5ca5d202b2074a1c` (before #163 )

### To Reproduce

**Expose TCP Ports**: `10101`

**Env variables**:
| Key | Value |
| --- | --- |
| H2O_LLM_STUDIO_WORKDIR | `/home/llmstudio/huggingface-cache/hub` |
| HF_DATASETS_CACHE | `/home/llmstudio/huggingface-cache/datasets` |
| HUGGINGFACE_HUB_CACHE | `/home/llmstudio/huggingface-cache/hub` |
| TRANSFORMERS_CACHE | `/home/llmstudio/huggingface-cache/hub` |

- [x] 🐛 `/workspace` is not writable

**To Reproduce / Runpod template**:

Use the following for all and override as instructed:
**Volume Mount Path**: `/workspace`
**Env vars**:
| Key | Value |
| --- | --- |
| H2O_LLM_STUDIO_WORKDIR | `/workspace` |

**Symptom**:

All of the files and directories in `/workspace` are owned by `root` while the user is `llmstudio`:

```
llmstudio@/workspace$ pwd
/workspace
llmstudio@/workspace$ ls -la
total 348
drwxr-xr-x 1 root root   4096 Jul 12 04:10 .
drwxr-xr-x 1 root root     50 Jul 13 01:29 ..
-rw-r--r-- 1 root root     25 Jul 12 04:02 .dockerignore
-rw-r--r-- 1 root root    308 Jul 12 04:02 .flake8
-rw-r--r-- 1 root root   2023 Jul 12 04:02 .gitignore
[...]

PR #285, opened 2023-07-17
Quantize models with AutoGPTQ (#219)

### LINKED ISSUE #219: [FEATURE] Option to Quantize / GPTQ first then push to Huggingface
### 🚀 Feature

See https://github.com/PanQiWei/AutoGPTQ/tree/main#quantization-and-inference

### Motivation

Much less VRAM & faster inference speeds.

Glavin001:
Close #219 

Reference code for successfully quantizing an LLM: https://gist.github.com/Glavin001/d66dbb93e9ea3a2cb5e0e6f1c2aa2ba1

Expects https://github.com/h2oai/h2o-llmstudio/issues/280 to be merged prior.

- [ ] Testing. Currently untested. Only had a CPU computer today. Will test with GPU later this week.
- [x] Add button to experiment's page
  - [ ] Only show button after model successfully trained (like the push, etc buttons)
- [ ] Add `autogptq` to Pipfile
- [ ] Sample from experiment's dataset
  - [x] Use experiment's configured prompt, answer, separator tokens
- [ ] Quantization configuration options per experiment
- [ ] Add AutoGPTQ inference support to Chat tab
- [ ] Push `GPTQ` model to Huggingface

### DIFF
--- a/app_utils/sections/experiment.py
+++ b/app_utils/sections/experiment.py
@@ -3,6 +3,7 @@ import logging
 import os
 import shutil
 import zipfile
+import time
 from typing import Callable, List, Optional, Set
 
 import accelerate
@@ -16,6 +17,10 @@ import yaml
 from h2o_wave import Q, data, ui
 from jinja2 import Environment, FileSystemLoader
 from sqlitedict import SqliteDict
+from auto_gptq import AutoGPTQForCausalLM, BaseQuantizeConfig
+
+# FIXME: remove?
+from datasets import load_dataset, Dataset
 
 from app_utils.config import default_cfg
 from app_utils.sections.chat import chat_tab, load_cfg_model_tokenizer
@@ -957,6 +962,11 @@ async def experiment_display(q: Q) -> None:
             name="experiment/display/download_logs",
             label="Download logs/config",
             primary=False,
+        ),
+        ui.button(
+            name="experiment/display/quantize_model",
+            label="Quantize model",
+            primary=False,
         )
     ]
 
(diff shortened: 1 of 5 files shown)


app_utils/sections/experiment.py
```diff
@@ -16,6 +17,10 @@
 from h2o_wave import Q, data, ui
 from jinja2 import Environment, FileSystemLoader
 from sqlitedict import SqliteDict
+from auto_gptq import AutoGPTQForCausalLM, BaseQuantizeConfig
+
+# FIXME: remove?
+from datasets import load_dataset, Dataset
```
Glavin001, line 23: May not need this import after getting `cfg.dataset` to work for extracting all of the examples

app_utils/sections/experiment.py
```diff
@@ -16,6 +17,10 @@
 from h2o_wave import Q, data, ui
 from jinja2 import Environment, FileSystemLoader
 from sqlitedict import SqliteDict
+from auto_gptq import AutoGPTQForCausalLM, BaseQuantizeConfig
```
Glavin001, line 20: Need to add to Pipfile as dependency

app_utils/sections/experiment.py
```diff
@@ -1471,6 +1481,121 @@ async def experiment_download_model(q: Q, error: str = ""):
     await q.page.save()
 
 
+async def experiment_quantize_model(q: Q, error: str = ""):
+    experiment = q.client["experiment/display/experiment"]
+    experiment_path = q.client["experiment/display/experiment_path"]
+    logger.info(f"Quantizing model!")
+    zip_path = get_model_path(experiment.name, experiment_path)
+
+    if not os.path.exists(zip_path):
```
Glavin001, line 1490: Check if the model is already quantized:

```suggestion
    if not os.path.exists(quantized_model_dir):
```

app_utils/sections/experiment.py
```diff
@@ -1471,6 +1481,121 @@ async def experiment_download_model(q: Q, error: str = ""):
...
+        if num_running_queued > 0:
+            logger.info("Waiting for queue to finish first.")
+            await q.page.save()
+            return
+
+        cfg, model, tokenizer = load_cfg_model_tokenizer(
+            experiment_path, merge=True, device=device
+        )
+
+        # Save model
```
Glavin001, line 1509: Not sure whether `save_pretrained` already detects when it was saved before, and skips fast?
Maybe wrap in a if-statement guard.

app_utils/sections/experiment.py
```diff
@@ -1471,6 +1481,121 @@ async def experiment_download_model(q: Q, error: str = ""):
...
+        model = unwrap_model(model)
+        checkpoint_path = cfg.output_directory
+        model.backbone.save_pretrained(checkpoint_path)
+        tokenizer.save_pretrained(checkpoint_path)
+
+        quantize_config = BaseQuantizeConfig(
+            bits=4,  # quantize model to 4-bit
+            group_size=128,  # it is recommended to set the value to 128
+            desc_act=False,  # set to False can significantly speed up inference but the perplexity may slightly bad
```
Glavin001, line 1518: Should this be configurable per experiment? Most I've seen always use this config.

app_utils/sections/experiment.py
```diff
@@ -1471,6 +1481,121 @@ async def experiment_download_model(q: Q, error: str = ""):
...
+
+                input_ids.append(tokenized_data["input_ids"][: tokenizer.model_max_length])
+                attention_mask.append(tokenized_data["attention_mask"][: tokenizer.model_max_length])
+
+            return {
+                "input_ids": input_ids,
+                "attention_mask": attention_mask,
+            }
+
+        # FIXME: Extract from real dataset
```
Glavin001, line 1547: Extract from real dataset

app_utils/sections/experiment.py
```diff
@@ -1471,6 +1481,121 @@ async def experiment_download_model(q: Q, error: str = ""):
...
+        raw_data = Dataset.from_dict({"instruction": [
+            "This is a test"
+        ], "input": [
+            "This is the answer"
+        ]})
+
+        n_samples = 128
```
Glavin001, line 1554: This should be configurable

app_utils/sections/experiment.py
```diff
@@ -1471,6 +1481,121 @@ async def experiment_download_model(q: Q, error: str = ""):
...
+        # Sample from the dataset if n_samples is provided and less than the dataset size
+        if n_samples is not None and n_samples < len(raw_data):
+            raw_data = raw_data.shuffle(seed=42).select(range(n_samples))
```
Glavin001, line 1557: Seed should be configurable

app_utils/sections/experiment.py
```diff
@@ -1471,6 +1481,121 @@ async def experiment_download_model(q: Q, error: str = ""):
...
+
+        logger.info("Quantize...")
+        start = time.time()
+        # quantize model, the examples should be list of dict whose keys can only be "input_ids" and "attention_mask"
+        qmodel.quantize(
+            examples_for_quant,
+            batch_size=1,
+            # batch_size=args.quant_batch_size,
+            # use_triton=args.use_triton,
+            # autotune_warmup_after_quantized=args.use_triton
```
Glavin001, line 1585: These should be configurable?

app_utils/sections/experiment.py
```diff
@@ -1471,6 +1481,121 @@ async def experiment_download_model(q: Q, error: str = ""):
...
+        )
+        end = time.time()
+        logger.info(f"quantization took: {end - start: .4f}s")
+
+        # save quantized model
+        logger.info("Saving quantized model...")
+        quantized_model_dir = checkpoint_path + "_quantized"
+        qmodel.save_quantized(quantized_model_dir, use_safetensors=True)
```
Glavin001, line 1593: Need a way to push after saving. Maybe do it in one step, as AutoGPT states is possible all together?

psinger:
Thanks a lot @Glavin001 for starting this!

I am curious, but is there an advantage of AutoGPT?
Couldnt we just use bitsandbytes like HF does?
https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L2709

Glavin001:
Asked the king of quantization, @TheBlokeAI in Discord: https://discord.com/channels/1111983596572520458/1117037259603066980/1130501685349978226

Highlights:
- Significantly smaller distribution storage size of weights. Distribute quantized not full weights.
- Pre-quantized, not as load time, so faster start
- GPTQ is more accurate when configured appropriately
- AutoGPTQ and especially ExLlama we’re faster — although new 4bit inference code in bitsandbytes may change that 

Full answers:

> A major advantage is that with GPTQ you only download the quantised weights. With bitsandbytes you have to download the full model weights, eg 13GB for 7B, 26GB for 13B, 61GB for 33B, 122GB for 65B
> 
> In terms of accuracy, GPTQ is I think a bit better when the best params are used, although that currently requires ExLlama for good performance.  Using parameters optimal for AutoGPTQ speed, accuracy will be around the same, comparing bitsandbytes to GPTQ
> 
> Speed I don't know. It used to be that AutoGPTQ was faster than bitsandbytes 4bit, and ExLlama was significantly faster
>
> But in the last few days, bitsandbytes 4bit apparently got a big speed boost. I've not been able to benchmark it yet, since that happened
>
> But IMHO the first difference I mentioned, the requirement to download full model weights for bitsandbytes, is going to be a deal breaker for most people
>
> AutoGPTQ is pre-quantised yes. You load the weights and save a quantized version of it, and the user loads that quantized version
>
> bitsandbytes quantises on the fly, hence the need for the full weights downloaded.  I don't know what the startup time for bitsandbytes is - probably a bit longer yes

psinger:
> A major advantage is that with GPTQ you only download the quantised weights. With bitsandbytes you have to download the full model weights, eg 13GB for 7B, 26GB for 13B, 61GB for 33B, 122GB for 65B

I dont think this is true. You can just push the quantized weights also with bitsandbytes/HF and then load them that way.

Does it need different code also for loading GPTQ weights?

Glavin001:
> You can just push the quantized weights also with bitsandbytes/HF

It appears so: https://huggingface.co/docs/transformers/main_classes/quantization#push-quantized-models-on-the-hub

```
model = AutoModelForCausalLM.from_pretrained("bigscience/bloom-560m", device_map="auto", load_in_8bit=True)

model.push_to_hub("bloom-560m-8bit")
```

> Pushing 8bit models on the Hub is strongely encouraged for large models. This will allow the community to benefit from the memory footprint reduction and loading for example large models on a Google Colab.

So LLM Studio could add support via `bitsandbytes` for exporting an already quantized models.
Previously this may have been only **practically** supported 8 bit, since the inference speed was so slow (from what I've read), however, [now that 4-bit inference is officially supported](https://github.com/TimDettmers/bitsandbytes/releases/tag/0.40.0) bitsandbytes is more competitive to GPTQ?


> Does it need different code also for loading GPTQ weights?

Yes, such as using [AutoGPTQ](https://github.com/PanQiWei/AutoGPTQ) or more popularly for LLaMA models, [ExLlama model loading](https://github.com/turboderp/exllama). ExLlama is much faster and the go-to right now for LLaMA based models and [depends on GPTQ, not bitsandbytes from my understanding](https://github.com/turboderp/exllama/issues/69#issuecomment-1596273647).

So the ask here may actually be: quantize and export via both bitsandbytes and GPTQ `.safetensors` for compatibility with the specific model loader which best suits your needs.

In my case, I specifically want ExLlama which needs GPTQ models in `.safetensors` format.

TheBloke:
Unfortunately you can't push 4bit models at this time.
```
NotImplementedError: You are calling `save_pretrained` on a 4-bit converted model. This is currently not supported
```

I guess they will hopefully add that in time, and yes it works with 8-bit. But then 8-bit is still very large for many users, and also much slower.

psinger:
Yes, 4bit will not work to push with bitsandbytes.

But pushing quantized weights with AutoGPTQ will always require to also use that for inference then, right? So cant keep staying only in HF eco-system afterwards.

Glavin001:
Appears so. Hopefully HF's solutions improve over time so the need is diminished. For now, alternatives are far more compelling for memory consumption and token/sec inference speed.

Recent news related to quantization and Llama-2: Bitsandbytes 4-bit is now comparable to AutoGPTQ / GPTQ-for-Llama, however, ExLlama which depends on GPTQ still leaves them all in the dust: https://twitter.com/abacaj/status/1681725266865864712?s=46&t=9HxibQPFoKlrGEEL9RigDA

image

psinger:
@Glavin001 I have been playing a bit with AutoGPTQ after your suggestion. 

I think it is fairly usable, but issue is that it can take quite some time to quantize. Im wondering if this will hinder the whole upload experience.

Maybe for starters support 8bit bitsandbytes upload?
But Im fine adding it, but would like to have both 8bit and 4bit options then.

There is also another PR for adding a script to upload
https://github.com/h2oai/h2o-llmstudio/pull/300

We could also add quantization options there.

PR #288, opened 2023-07-18
deepspeed

haqishen:
## Feature

* deepspeed zero3 training
* deepspeed zero3 /w lora training
* deepspeed zero3 /w offload optimizer training

### Full params experiments

| -- | backbone | dtype | LORA | deepspeed | offload optimizer (live param 1e9) | RAM usage per GPU | runtime(data sample 0.1) | Perplexity |
| -------- | -------- | ----- | ---- | --------- | ----------------- | ----------------- | ------------------------ | ---------- |
| exp1 | EleutherAI/pythia-1b-deduped | bfloat16 | False | False | False | 15GB | 00:03:28 | 11.9258 |
| exp2 | EleutherAI/pythia-1b-deduped (lr 0.00001) | float16 | False | **True** | False | 11.5GB | 00:04:31 | 10.1721 |
| exp3 | EleutherAI/pythia-1b-deduped (lr 0.00001) | float16 | False | **True** | **True** | 5GB | 00:29:07 | 10.1965 |
| exp4 | EleutherAI/pythia-2.8b-deduped | bfloat16 | False | False | False | OOM | N/A | N/A |
| exp5 | EleutherAI/pythia-2.8b-deduped | float16 | False | **True** | False | OOM | N/A | N/A |
| exp6 | EleutherAI/pythia-2.8b-deduped | float16 | False | **True** | **True** | 10.5G | 01:18:44 | 8.7803 |
| exp7| EleutherAI/pythia-6.9b-deduped | float16 | False | **True** | **True** (live param 1e10) | 23G | OOM (cpu) | OOM (cpu) |

### LORA experiments

| -- | backbone | dtype | LORA | deepspeed | RAM usage per GPU | runtime(data sample 0.1) | Perplexity |
| -------- | -------- | ----- | ---- | --------- | ----------------- | ------------------------ | ---------- |
| exp1 | EleutherAI/pythia-2.8b-deduped | float16 | **True** | False | 11.5GB | 00:00:57 | 9.7002 |
| exp2 | EleutherAI/pythia-2.8b-deduped | float16 | **True** | **True** | 4.5GB | 00:08:19 | 9.8184 |
| exp3 | EleutherAI/pythia-6.9b-deduped | float16 | **True** | False | 16.5GB | 00:01:40 | 9.4829 |
| exp4 | EleutherAI/pythia-6.9b-deduped | float16 | **True** | **True** | 8.5GB | 00:23:35 | 9.6707 |
| exp5 | EleutherAI/pythia-12b-deduped | float16 | **True** | False | OOM | NA | NA |
| exp6 | EleutherAI/pythia-12b-deduped | float16 | **True** | **True** | 12.5GB | 00:46:32 | 9.1973 |
| exp7 | EleutherAI/pythia-12b-deduped | int8 | **True** | False | 17GB | 00:06:58 | 10.1232 |
| exp8 | EleutherAI/pythia-20b-deduped | float16 | **True** | **True** | 18.5GB | 00:56:51 | 8.4031 |

#### Using 8 x V100 w/ NVLink (16GB), Batchsize = 1

| -- | backbone | dtype | LORA | deepspeed | RAM usage per GPU | runtime(data sample 0.1) | Perplexity |
| -------- | -------- | ----- | ---- | --------- | ----------------- | ------------------------ | ---------- |
| exp1 | EleutherAI/pythia-20b-deduped | int4 | **True** | **False** | 15.5GB | 00:02:29 | 9.3201 |
| exp2| EleutherAI/pythia-20b-deduped | float16 | **True** | **True** | 10.5GB | 00:04:19 | 8.7182 |

#### Using 8 x A6000 w/o NVLink (48GB), Batchsize = 1

| -- | backbone | dtype | LORA | deepspeed | RAM usage per GPU | runtime(data sample 0.1) | Perplexity |
| -------- | -------- | ----- | ---- | --------- | ----------------- | ------------------------ | ---------- |
| exp1 | tiiuae/falcon-40b | int4 | **True** | False | 45GB | 00:12:41 | 5.7722 |
| exp2 | tiiuae/falcon-40b | float16 | **True** | **True** | 22GB | 02:30:52 | 5.7743 |
| exp3 | TheBloke/Llama-2-70B-Chat-fp16 | int4 | **True** | False | 46GB | 00:20:56 | 4.4524 |
| exp4 | TheBloke/Llama-2-70B-Chat-fp16 | float16 | **True** | **True** | 28GB | 04:30:58 | 4.4221 |

NVLINK works:

| -- | backbone | dtype | LORA | deepspeed | RAM usage per GPU | runtime(data sample 0.1) |
| -------- | -------- | ----- | ---- | --------- | ----------------- | ------------------------ | 
| w/ NVLINK | EleutherAI/pythia-20b-deduped | float16 | **True** | **True** | 10.5GB | 00:04:19 | 
| w/o NVLINK | EleutherAI/pythia-20b-deduped | float16 | **True** | **True** | 10.5GB | 00:45:37 | 

#### Using 8 x A100 SMX4 

| -- | backbone | dtype | LORA | deepspeed | RAM usage per GPU | runtime |
| -------- | -------- | ----- | ---- | --------- | ----------------- | ------------------------ | 
| exp1 | TheBloke/Llama-2-70B-Chat-fp16 (4k) | int4 | **True** | False | ~80GB | 35h | 
| exp2 | TheBloke/Llama-2-70B-Chat-fp16 (4k) | float16 | **True** | **True** |  ~80GB | 6.5h | 
| exp3 | h2oai/h2ogpt-4096-llama2-13b-chat | float16 | **True** | **True** |  11GB | 16min| 
| exp4 | h2oai/h2ogpt-4096-llama2-13b-chat | float16 | **True** | **False** |  38GB | 13min | 

## Check

- [x] Chat
- [x] Upload model weight

## Future Work

* zero ++
* zero3 /w lora and offload optimizer
* zero3 /w offload params


llm_studio/src/utils/modeling_utils.py
```diff
@@ -214,59 +271,66 @@ def get_optimizer(model: torch.nn.Module, cfg: Any) -> torch.optim.Optimizer:
...
-                ],
-                "lr": cfg.training.differential_learning_rate,
-                "weight_decay": 0,
-            },
-        ],
-        lr=cfg.training.learning_rate,
-        weight_decay=cfg.training.weight_decay,
-    )
+    if cfg.environment.use_deepspeed and cfg.training.lora:
+        logger.info(
```
haqishen, line 275: Maybe deepspeed support differential learning rate. Just not support this implementation. I'm Not sure.

llm_studio/src/utils/modeling_utils.py
```diff
@@ -169,6 +170,65 @@ def load_checkpoint(
...
+            "enabled": True if cfg.architecture.backbone_dtype == "float16" else False
+        },
+        "bf16": {
+            "enabled": True if cfg.architecture.backbone_dtype == "bfloat16" else False
+        },
+        # https://www.deepspeed.ai/docs/config-json/#zero-optimizations-for-fp16-training
+        "zero_force_ds_cpu_optimizer": False,
+        "zero_optimization": {
+            "stage": 3,
+            "mics_shard_size": cfg.environment._world_size,
```
haqishen, line 199: By adding this line, GPU RAM usage per GPU go down from 11.5GB to 8.5GB

train.py
```diff
@@ -442,25 +469,27 @@ def run_train(
...
+                else:
+                    checkpoint_path = cfg.output_directory
+                    if cfg.environment._local_rank == 0:
                         logger.info(
                             f"Saving last model checkpoint: "
                             f"val_loss {val_loss:.5}, val_{cfg.prediction.metric} "
                             f"{val_metric:.5} to {checkpoint_path}"
                         )
-                        save_checkpoint(model=model, path=checkpoint_path, cfg=cfg)
+                    save_checkpoint(model=model, path=checkpoint_path, cfg=cfg)
```
pascal-pfeiffer, line 492: why did this change?

`save_checkpoint` is called on all ranks?

train.py
```diff
@@ -489,6 +518,10 @@ def run(cfg: Any) -> None:
         cfg: config object with all the hyperparameters
     """
 
+    if cfg.environment.use_deepspeed:
+        # To avoid warnings about parallelism in tokenizers
+        os.environ["TOKENIZERS_PARALLELISM"] = "false"
+
```
pascal-pfeiffer, line 524: should already be set on the very top of this file. Why again?

train.py
```diff
@@ -625,10 +661,11 @@ def run(cfg: Any) -> None:
                     param.requires_grad = True
                     param.data = param.data.float()
 
-    if cfg.environment._distributed:
+    if cfg.environment._distributed and not cfg.environment.use_deepspeed:
         model = wrap_model_distributed(model, cfg, cfg.environment.use_fsdp)
 
-    if cfg.environment.compile_model:
+    # deepspeed do not support torch.compile
+    if cfg.environment.compile_model and not cfg.environment.use_deepspeed:
```
pascal-pfeiffer, line 668: throw a warning, if the compile was selected with deepspeed

train.py
```diff
@@ -178,6 +181,22 @@ def run_train(
...
+            scheduler,
+        ) = deepspeed_initialize(
+            model=model,
+            optimizer=optimizer,
+            lr_scheduler=scheduler,
+            training_data=train_dataloader.dataset,
+            validating_data=val_dataloader.dataset,
+            cfg=cfg,
+        )
+
```
pascal-pfeiffer, line 199: should that maybe go to the `wrap_model_distributed` function to reduce if/else checks?

llm_studio/src/utils/modeling_utils.py
```diff
@@ -214,59 +301,66 @@ def get_optimizer(model: torch.nn.Module, cfg: Any) -> torch.optim.Optimizer:
...
-    )
+    if cfg.environment.use_deepspeed and cfg.training.lora:
+        logger.info(
+            "Deepspeed /w Lora training do not support differential learning rate."
+        )
+        optimizer = Optimizers.get(cfg.training.optimizer)(
+            model.parameters(), lr=cfg.training.learning_rate
+        )
+    else:
+        no_decay = ["bias", "LayerNorm.weight"]
```
pascal-pfeiffer, line 312: how is `no_decay = ["bias", "LayerNorm.weight"]` handled with deepspeed?

if different, this needs documentation and warnings

llm_studio/src/utils/modeling_utils.py
```diff
@@ -169,6 +183,79 @@ def load_checkpoint(
...
+    ds_config = {
+        "fp16": {
+            "enabled": True if cfg.architecture.backbone_dtype == "float16" else False,
+            "loss_scale_window": 100,
+        },
+        "bf16": {
+            "enabled": True if cfg.architecture.backbone_dtype == "bfloat16" else False,
+            "loss_scale_window": 100,
+        },
+        # https://www.deepspeed.ai/docs/config-json/#zero-optimizations-for-fp16-training
```
pascal-pfeiffer, line 210: we should probably throw a warning for int4 and int8 selections and force to float16

llm_studio/src/utils/modeling_utils.py
```diff
@@ -75,15 +77,27 @@ def save_checkpoint(model: torch.nn.Module, path: str, cfg: Any):
...
+            model.save_checkpoint(os.path.join(path, "ds_checkpoint"))
+            if cfg.environment._local_rank == 0:
+                # load to cpu
+                state_dict = get_fp32_state_dict_from_zero_checkpoint(
+                    os.path.join(path, "ds_checkpoint")
+                )
+                # save as normal checkpoint that can be loaded by `load_state_dict`
+                checkpoint = {"model": state_dict}
+                torch.save(checkpoint, os.path.join(path, "checkpoint.pth"))
+                shutil.rmtree(os.path.join(path, "ds_checkpoint"))
```
pascal-pfeiffer, line 92: where are we saving the LoRA adapter?

pascal-pfeiffer (dismissed):
Thanks a lot for working on this tricky topic, @haqishen 

I think we can remove offloading to CPU for the time being, as the speed apparently drops way too much. 

I already added a few comments, will need to do some proper testing on a multi GPU setup and also with continued training, upload of weights, etc. I assume, you checked these already?

train.py
haqishen: Sorry for the late, I didn't realize you already reviewing this PR.

For deepspeed, saving checkpoint need to be called on all ranks, so I move the `if cfg.environment._local_rank == 0:` into function `save_checkpoint` for DDP.

train.py
haqishen: [d30b51c](https://github.com/h2oai/h2o-llmstudio/pull/288/commits/d30b51c36d7dc3de5814bc561db826c196ce615b)

fixed.

train.py
haqishen: fixed

[67629ee](https://github.com/h2oai/h2o-llmstudio/pull/288/commits/67629eee9dd873992b035626d0fe2ee9f23f7080)

llm_studio/src/utils/modeling_utils.py
haqishen: [d1efef5](https://github.com/h2oai/h2o-llmstudio/pull/288/commits/d1efef56fd2a7de4de01dff4a1b511696294803c)

Raise error for better understanding they are using FP16 when deepspeed.

train.py
haqishen: [d1efef5](https://github.com/h2oai/h2o-llmstudio/pull/288/commits/d1efef56fd2a7de4de01dff4a1b511696294803c)

update here

llm_studio/src/utils/modeling_utils.py
haqishen: I dont' have good idea to save it as lora.
I tried to do the following but get stuck when creating model_for_save.
Do you have any idea on it? @pascal-pfeiffer 

image

llm_studio/src/utils/modeling_utils.py
haqishen: If I uncomment those `and param.requires_grad`
in this function it would work for deepspeed + lora as well.
Is there any special reason for commenting out `and param.requires_grad` ?

pascal-pfeiffer: Well, we don't apply weight decay to `["bias", "LayerNorm.weight"]`, that does make a difference.

llm_studio/src/utils/modeling_utils.py
haqishen: fix in [7b78538](https://github.com/h2oai/h2o-llmstudio/pull/288/commits/7b7853868f19d6e483462e20b5c14f0767535415)

llm_studio/src/utils/modeling_utils.py
haqishen: Deepspeed do not has special control on `["bias", "LayerNorm.weight"]`

I fix it as the following commit

[892f47c](https://github.com/h2oai/h2o-llmstudio/pull/288/commits/892f47cbfe6c56af96ffae737d5234fc825db332)

train.py
```diff
@@ -350,25 +356,26 @@ def run_train(
...
+                else:
+                    checkpoint_path = cfg.output_directory
+                    if cfg.environment._local_rank == 0:
                         logger.info(
                             f"Saving last model checkpoint: "
                             f"val_loss {val_loss:.5}, val_{cfg.prediction.metric} "
                             f"{val_metric:.5} to {checkpoint_path}"
                         )
-                        save_checkpoint(model=model, path=checkpoint_path, cfg=cfg)
+                    save_checkpoint(model=model, path=checkpoint_path, cfg=cfg)
```
pascal-pfeiffer, line 378: Does it save on all ranks now? Why would that be needed and does it not conflict with standard DDP?

hatrexltd:
During training with the Llama 70B 2 model with deepspeed, in the evaluations the results are inconsistent and have no relation to the question.

train.py
```diff
@@ -278,7 +281,7 @@ def run_train(
                 loss = loss / cfg.training.grad_accumulation
 
             # Backward pass
-            if cfg.environment.mixed_precision:
+            if cfg.environment.mixed_precision and not cfg.environment.use_deepspeed:
```
psinger, line 284: How exactly does deepspeed handle mixed precision training?

psinger (changes requested):
Thanks @haqishen for the great work. Let's have hpefully one (max two) final minor round and then I believe this should be good to be merged in a first stage. Please then also open new issue(s) for next potential improvements and we can fix other issues that emerge along the way.

Some questions:

- Do I read the table right, that after NVLINK / SMX4 the runtime goes down a lot with deepspeed? Does it match ddp runtime then?

In this table deepspeed is even significantly faster? Is this an effect of deepspeed or float16?
image

- How exactly does the sharding work with LORA? Are all layers sharded? Would it be useful/possible to keep the LORA layers on a single gpu?

Here are some notes:

Val order still seems wrong?
image
Might be related to my inline comment

I believe the parameter size calculation is somehow off. When I am running b13 llama2 in ddp mode, I get `26044835840` bytes. With deepspeed I get `26214400`. From my understanding deepspeed should be twice the size, not much smaller.

image
Can we make these sliders?

image
Let's fully remove FSDP in favor of Deepspeed

llm_studio/src/utils/modeling_utils.py
```diff
@@ -288,10 +290,13 @@ def wrap_model_distributed(
...
-            model=torch.nn.Linear(1, 1),
-            training_data=val_dataloader.dataset,
-            config_params=ds_config,
+        val_dataloader = DeepSpeedDataLoader(
+            val_dataloader.dataset,
+            batch_size=val_dataloader.batch_size,
+            local_rank=cfg.environment._local_rank,
+            pin_memory=True,
+            tput_timer=None,
+            data_sampler=DistributedSampler(val_dataloader.dataset, shuffle=False),
```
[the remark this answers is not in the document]
psinger: @haqishen I pushed a change there, please make fully sure it works accordingly for validation order

llm_studio/src/utils/modeling_utils.py
```diff
@@ -90,15 +97,29 @@ def save_checkpoint(model: torch.nn.Module, path: str, cfg: Any):
...
-        model.backbone.save_pretrained(path)
-
-    checkpoint = {"model": model.state_dict()}
-
-    if path is not None:
-        torch.save(checkpoint, os.path.join(path, "checkpoint.pth"))
+    if cfg.environment.use_deepspeed:
+        if path is not None:
+            # gather model params from all ranks
+            if hasattr(cfg.training, "lora") and cfg.training.lora:
```
psinger, line 103: should this only be for rank 0?

llm_studio/src/utils/modeling_utils.py
```diff
@@ -90,15 +97,29 @@ def save_checkpoint(model: torch.nn.Module, path: str, cfg: Any):
...
+                model.backbone.save_pretrained(path)
+            model.save_checkpoint(os.path.join(path, "ds_checkpoint"))
```
psinger, line 105: also here? seems there might be desync issues

llm_studio/src/utils/modeling_utils.py
```diff
@@ -90,15 +97,29 @@ def save_checkpoint(model: torch.nn.Module, path: str, cfg: Any):
...
+            if cfg.environment._local_rank == 0:
+                # load to cpu
+                state_dict = get_fp32_state_dict_from_zero_checkpoint(
```
psinger, line 108: and why do we need this separately?

psinger: or rather, shouldnt that be sufficient?

Pipfile
```diff
@@ -49,6 +49,7 @@ h2o-wave = "0.26"
 tiktoken = "0.4.0"
 hf-transfer = "0.1.3"
 peft = "0.4.0"
+deepspeed = ">=0.10.0"
```
psinger, line 52: fix version

psinger:
One more find, could this be useful for saving?
https://deepspeed.readthedocs.io/en/stable/zero3.html#deepspeed.runtime.zero.config.DeepSpeedZeroConfig.gather_16bit_weights_on_model_save

llm_studio/src/utils/modeling_utils.py
haqishen: For deepspeed, saving checkpoint requires to be called in all ranks for gathering every parameters of the model.

psinger: also lora?

haqishen: > How exactly does the sharding work with LORA? Are all layers sharded? Would it be useful/possible to keep the LORA layers on a single gpu?

The method of sharding is no different in lora training and non-lora training. Moreover, I believe there is no need to make all lora layers on a single GPU, otherwise it would make that GPU the bottleneck of the entire training.

Therefore, when saving the model, regardless of whether it's lora training or not, it's necessary to call save_checkpoint on all ranks.

llm_studio/src/utils/modeling_utils.py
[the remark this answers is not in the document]
haqishen: Thanks for fixing this! I can confirm that the order of validation is now correct.

train.py
haqishen: When use deepspeed and type fp16 or bf16, it will do mixed precision automatically.

https://huggingface.co/docs/transformers/main_classes/deepspeed#automatic-mixed-precision

Maybe we can warning it and force mixed precision to be True when using deepspeed?

llm_studio/python_configs/text_causal_language_modeling_config.py
```diff
@@ -343,6 +343,13 @@ class ConfigNLPCausalLMEnvironment(DefaultConfig):
 
     compile_model: bool = False
     use_fsdp: bool = False
+    use_deepspeed: bool = False
+    deepspeed_reduce_bucket_size: int = 1e6
+    deepspeed_stage3_prefetch_bucket_size: int = 1e6
+    deepspeed_stage3_param_persistence_threshold: int = 1e6
+    #     deepspeed_offload_optimizer: bool = False
```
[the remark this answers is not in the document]
haqishen: In case we need to implement the 'offload to CPU' feature in the future, keep the relevant code.

haqishen:
> Do I read the table right, that after NVLINK / SMX4 the runtime goes down a lot with deepspeed? Does it match ddp runtime then?

Just updated the table by adding some exp result to compare ddp and deepspeed runtime. It shows that deepspeed still slower than ddp fp16 by around 15~20%, but much faster than ddp int8 or int4.

> Can we make these sliders?

How to do this? I search the keyword `data_sample` but cannot find why it's a slider bar in webui.

> Let's fully remove FSDP in favor of Deepspeed

Let's do it in a new PR.

psinger:
@haqishen I believe we still have not solved the desync issue, when long-running generate is running, and then the checkpoint is saved afterwards. I believe we should move the order there as discussed earlier. So save the checkpoint before running eval, if not best epoch saving is enabled.

haqishen:
> @haqishen I believe we still have not solved the desync issue, when long-running generate is running, and then the checkpoint is saved afterwards. I believe we should move the order there as discussed earlier. So save the checkpoint before running eval, if not best epoch saving is enabled.

btw what's your experiment setting?

psinger:
> > @haqishen I believe we still have not solved the desync issue, when long-running generate is running, and then the checkpoint is saved afterwards. I believe we should move the order there as discussed earlier. So save the checkpoint before running eval, if not best epoch saving is enabled.
> 
> btw what's your experiment setting?

default with gpt metric

psinger (approved):
I fixed another issue.

Need to merge main, solve conflicts and then can merge.

After that, please open subsequent issues for things not tackled yet in this pr and potential future improvements.

We also potentiall need a section in README / Docs and might be also useful to share your benchmarks there.

Issue #283, opened 2023-07-14
[CODE IMPROVEMENT] Prevent chained conversations being stuck in infinite loop


psinger:
### 🔧 Proposed code refactoring

If there is a bug in `parent_id` and `id` structure, it can be stuck in an infinite loop.

We should have some safeguarding against that, either some initial checks, or max iterations.

PR #289, opened 2023-07-18
Add data sanity checks

### LINKED ISSUE #257: [FEATURE] Prevent infinite loop when selecting row id as "parent_id"
_###_ 🐛 Bug

I am getting a Data Error while running an experiment on LLM Studio:

**ERROR: Data error occurred during H2O LLM Studio run**:
_Traceback (most recent call last):_
_File "/root/.local/share/virtualenvs/h2o_-_llmstudio_-_29knupLy/lib/python3.10/site_-_packages/torch/utils/data/dataloader.py", line 1133, in_ _try_get_data
data = self._data_queue.get(timeout=timeout)
File "/usr/lib/python3.10/queue.py", line 180, in get
self.not_empty.wait(remaining)_

The experiment runs for hours with no progress and eventually times out.

### To Reproduce

Use a dataset other than 'oasst' and merely run an experiment.
(I've shared my sample dataset via email)

maxjeblick:
This PR adds sanity checks to the data to catch an error that occurs when `parent id == id` is set. To further avoid an infinite while loop originating from bad data input, the loop is stopped after 1000 iterations.

During refactoring I also changed the following two items. Happy to discuss :)

- `prompt_input_ids` is now truncated/padded to `cfg.tokenizer.max_length` instead of `cfg.tokenizer.max_length_prompt`.
I think the change is useful, as `max_length_prompt` denotes the length of an individual prompt, whereas `prompt_input_ids` may consist of a chained Q&A pairs where only the last answer has been removed. 

- ~~`cfg.augmentation.random_parent_probability` (add a random parent) is only used when the dataset has a parent id column.~~

Fixes https://github.com/h2oai/h2o-llmstudio/issues/283
Fixes #257

### DIFF
--- a/llm_studio/src/datasets/text_causal_language_modeling_ds.py
+++ b/llm_studio/src/datasets/text_causal_language_modeling_ds.py
@@ -467,21 +467,20 @@ class CustomDataset(Dataset):
         return [system_encoding, prompt_encoding, answer_encoding]
 
     def get_parent_ids(self, idx):
-        max_loop = 1000
+        max_loop = 1_000
         parent_idxs = []
         if self.parent_ids is not None:
             parent_idx = idx
             while (
                 (parent_idx := self.df_id_to_idx.get(self.parent_ids[parent_idx], None))
-                and max_loop > 0
             ) is not None:
                 parent_idxs.append(parent_idx)
                 max_loop -= 1
-
-        if max_loop == 0:
-            raise ValueError(
-                f"Parent chain of sample with idx {idx} exceeds max loop count."
-            )
+                if max_loop == 0:
+                    raise ValueError(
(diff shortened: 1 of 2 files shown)


psinger:
Any specific reason to not allow `cfg.augmentation.random_parent_probability` without `parent_id`?

I think my idea was that it could be also useful to augment with non-related messages for chat use cases.

maxjeblick:
> I think my idea was that it could be also useful to augment with non-related messages for chat use cases.

That's a good point, I actually noted that I already undid the corresponding change in the PR.

pascal-pfeiffer (approved):
thanks a lot @maxjeblick 
lgtm!
and thank you for adding the unit tests.

please link the corresponding issue before merging: https://github.com/h2oai/h2o-llmstudio/issues/283

PR #291, opened 2023-07-19
Package upgrades and minor changes

psinger:
In this PR I add the following things:

- update peft and transformers library
- add hf_transfer and `HF_HUB_ENABLE_HF_TRANSFER` for faster download/upload only for `make wave` - can be instable, so mostly for dev purposes
- minor change to properly load mpt models, they will run with `batch_size=1` for now


Pipfile
```diff
@@ -49,6 +49,7 @@ Jinja2 = ">=3.1.2, <4.0.0"
 tenacity = ">=8.2.2, <9.0.0"
 h2o-wave = "0.26"
 tiktoken = "*"
+hf-transfer = "*"
```
maxjeblick: Can we specify a version?

maxjeblick (approved):
Thanks for the changes LGTM! 
I left a small comment.

Pipfile
psinger: yes sorry I already pushed it in meantime

PR #292, opened 2023-07-19
Removal of llama force slow

psinger:
Decided to remove the manual llama slow tokenizer fix.

User should make this choice. LLama 2 for example works with fast tokenizer also.


pascal-pfeiffer (approved):
thank you, i was also wondering if we should remove it now

lgtm

PR #293, opened 2023-07-19
Refactor problem type

maxjeblick:
This PR refactors some parts of the code that are related to handling the problem configuration.

It should now be possible to add new problem types by creating a new configuration file with a corresponding `ConfigProblemBase` class.

### DIFF
--- a/llm_studio/src/datasets/text_utils.py
+++ b/llm_studio/src/datasets/text_utils.py
@@ -56,7 +56,10 @@ def get_tokenizer(cfg: Any):
         tokenizer.eos_token = "</s>"
 
     if tokenizer.pad_token is None:
-        tokenizer.pad_token = tokenizer.eos_token
+        if tokenizer.unk_token is not None:
+            tokenizer.pad_token = tokenizer.unk_token
+        else:
+            tokenizer.pad_token = tokenizer.eos_token
     if tokenizer.bos_token is None:
         tokenizer.bos_token = tokenizer.eos_token
     if tokenizer.cls_token is None:


llm_studio/src/utils/config_utils.py
```diff
@@ -87,7 +75,7 @@ def _get_type_annotation_error(v: Any, type_annotation: Type) -> ValueError:
     )
 
 
-def convert_cfg_to_nested_dictionary(cfg: ConfigProblemBase) -> dict:
+def convert_cfg_to_nested_dictionary(cfg) -> dict:
```
psinger, line 78: shall we still have `Any` type annotation there?

llm_studio/src/utils/config_utils.py
```diff
@@ -208,25 +200,12 @@ def load_config_yaml(path: str) -> ConfigProblemBase:
...
-        prediction=ConfigNLPCausalLMPrediction.from_dict(
-            cfg_dict.get("prediction", {})
-        ),
-        environment=ConfigNLPCausalLMEnvironment.from_dict(
-            cfg_dict.get("environment", {})
-        ),
-        logging=ConfigNLPCausalLMLogging.from_dict(cfg_dict.get("logging", {})),
-    )
-    return cfg
+    if cfg_dict["problem_type"] == "text_causal_language_modeling":
```
psinger, line 203: Shall we somewhere note that this needs to be added for new problem types? But probably it will be mostly us and we will hopefully remember or notice.

psinger (approved):
Thanks @maxjeblick - I believe all lgtm

llm_studio/src/utils/config_utils.py
maxjeblick: Good point! One could think about rewriting this or add better documentation. I'd propose to leave it as is for now and potentially refactor once we add more problem types.

llm_studio/src/utils/config_utils.py
maxjeblick: Not sure, Any is rather generic type. 
I may have a look into better type annotations when using config classes, which is currently not possible due to cyclic imports.

PR #294, opened 2023-07-19
LLama2 config fix

psinger:
This fix is needed to make LLama training currently work with peft and is also a reasonable change without lora training.

> Setting config.pretraining_tp to a value different than 1 will activate the more accurate but slower computation of the linear layers, which should better match the original logits.

See discussions:
https://github.com/huggingface/peft/pull/728
https://github.com/huggingface/transformers/pull/24906

### DIFF
--- a/llm_studio/src/utils/modeling_utils.py
+++ b/llm_studio/src/utils/modeling_utils.py
@@ -496,6 +496,7 @@ def update_backbone_config(config: Any, cfg: Any):
     if "mpt-" in cfg.llm_backbone:
         config.init_device = cfg.environment._device
 
+    # See: https://github.com/huggingface/transformers/pull/24906
     if hasattr(config, "pretraining_tp"):
         logger.info("Setting pretraining_tp of model config to 1.")
         config.pretraining_tp = 1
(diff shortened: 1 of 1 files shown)


pascal-pfeiffer (approved):
lgtm

could you add an in code comment to the discussion that you linked here, too?

llm_studio/src/utils/modeling_utils.py
```diff
@@ -496,6 +496,10 @@ def update_backbone_config(config: Any, cfg: Any):
     if "mpt-" in cfg.llm_backbone:
         config.init_device = cfg.environment._device
 
+    if hasattr(config, "pretraining_tp"):
+        logger.info("Setting pretraining_tp of model config to 1.")
```
pascal-pfeiffer, line 500: Add link to the discussion in the warning.

psinger: Dont think thats needed, it is a pretty safe switch
llama2-7b has it even as default

psinger: ok added comment and switch for lora

PR #296, opened 2023-07-19
Set huggingface token when loading tokenizers

### LINKED ISSUE #295: [BUG] Chatbot card broken for llama v2
### 🐛 Bug
Hugginface token is not set when loading the model in the app.
Thus, chat for llama v2 models (and probably other private models) fails with

`requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://huggingface.co/meta-llama/Llama-2-7b-chat-hf/resolve/main/tokenizer_config.json`

maxjeblick:
Fixes #295

### DIFF
--- a/app_utils/sections/chat.py
+++ b/app_utils/sections/chat.py
@@ -14,7 +14,7 @@ from h2o_wave import data as chat_data
 from h2o_wave import ui
 from transformers import AutoTokenizer, TextStreamer
 
-from app_utils.utils import get_experiments, get_ui_elements, parse_ui_elements
+from app_utils.utils import get_experiments, get_ui_elements, parse_ui_elements, set_env
 from llm_studio.src.datasets.text_utils import get_tokenizer
 from llm_studio.src.models.text_causal_language_modeling_model import Model
 from llm_studio.src.utils.config_utils import load_config_yaml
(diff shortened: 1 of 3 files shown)


pascal-pfeiffer (commented):
thank you. Max

could you add the same when calling  `create_nlp_backbone`, maybe?

maxjeblick:
> could you add the same when calling `create_nlp_backbone`, maybe?

Do we call this function somewhere within the app outside `load_cfg_model_tokenizer`?

pascal-pfeiffer:
> Do we call this function somewhere within the app outside load_cfg_model_tokenizer?

it's the function that we call when initialising the model

maxjeblick:
`load_cfg_model_tokenizer` will call `create_nlp_backbone` eventually, so that should be covered.

maxjeblick:
For `train.py`, env variable is set when starting from the app, for cli usage you would need to set the env variable `HUGGINGFACE_TOKEN` explicitly.

pascal-pfeiffer:
> For `train.py`, env variable is set when starting from the app, for cli usage you would need to set the env variable `HUGGINGFACE_TOKEN` explicitly.

right. That's where I missed it.

pascal-pfeiffer (approved):
thank you, lgtm!

Issue #88, opened 2023-05-03
[CODE IMPROVEMENT] Rework plotting


psinger:
The current plotting functionality could need some rework, some suggestions:

- For training plotting, I would just plot the whole sample, and (color)mark the labels.
- For validation insights we need better visual depictions of the different parts and clearer separation between samples.

psinger:
Addressed via https://github.com/h2oai/h2o-llmstudio/pull/301

PR #301, opened 2023-07-20
Max/insights table view

### LINKED ISSUE #299: [CODE IMPROVEMENT] Improve Validation Prediction Insights

Replace validation prediction insights by a sortable/searchable wave table.
In addition, consider switching from tooltip view to wrap view in sample train data visualization (see screenshot).

### Motivation

Validation prediction insights fit a table format naturally. By this, we also don't need to limit to 10 samples per random/best/worst tab. 

Screenshot 2023-07-20 at 11-26-22 H2O LLM Studio

maxjeblick:
This PR switches from markdown to a wave table view in validation prediction insights. The table can be sorted by metric value and is searchable.

Previous experiments will still visualize the markdown version of the insights, ensuring backwards compatibility.

Fixes #299

Screenshot 2023-07-20 at 14-19-02 H2O LLM Studio


psinger (commented):
Thanks @maxjeblick - looks and works awesome.

Partially also closes https://github.com/h2oai/h2o-llmstudio/issues/88

Would there be any way to render newlines? Probably not, right? Makes it a bit hard to judge some quality of the outputs, specifically code and lists.

A few more notes / questions:
- Is it also possible to do it similarly for train insights?
- Would it be possible to render code/markdown code?

psinger (commented):
@maxjeblick I saw you also discussed here in this issue:
https://github.com/h2oai/wave/discussions/2105

Now after seeing it - I think table view specifically for train insights is indeed a bit tricky to navigate:
image

For me personally, training insights is explicitly important to see the detailed formatting of the input, so every newline counts. Here in this case, I cant even scroll fully to the right very easily to see the linebreaks.

But what is pretty useful, is to hover over it:
image

Some general suggestions for train insights:
- I think showing both text and target text separately is a bit a waste of space. Do you think it would be easy to just show text, and then color-code the answer/loss part with a color? We should have the mask for it. Then it would be just two columns instead of four.
- shall we maybe hide the continuous left pads? Maybe just show one - would also save a lot of space.

Also for val insights, the hover render looks better than the actual table:
image

I will leave the decision to you of what format to use for train insights, I think everything is fine. The old one also was not ideal.
Maybe dont spend too much time on this. I can take another look, but please feel free to go with anything that you feel is a good start.

maxjeblick:
Thanks for the helpful feedback!
I've restructured the train data insights tab, it now shows prompt text, answer and tokenized text. The tokenized text has different formatting options to highlight pad tokens/prompt tokens and answer tokens.

I agree that the current table view is not optimal, but I wasn't able to implement a better solution quickly.

Screenshot 2023-07-31 at 14-01-13 H2O LLM Studio

maxjeblick:
I decided to transpose the wave table; the insights are more similar to the previous design and better readable.

Screenshot 2023-07-31 at 18-47-07 H2O LLM Studio
Screenshot 2023-07-31 at 18-47-18 H2O LLM Studio

fatihozturkh2o:
> I decided to transpose the wave table; the insights are more similar to the previous design and better readable.
> 
> Many thanks for @fatihozturkh2o for helpful brainstorming!

Definitely looks way better! 

2 quick suggestions from my side: 
1 - Maybe you can add a Sample Number column at the very left of the table, instead of adding " X Sample N" in the description column in each row. Seeing only sample numbers per row might visually help group rows more easily.

2 - In the validation insights tab, the "**metrics**" tab only shows numbers, so we don't have an idea about what the metric actually is. Appending the metric name to the same column could be a solution, off the top of my head. Like this ->  **"metric (BLEU or GPT..)**"

image

maxjeblick:
@fatihozturkh2o Thanks for the comments, I changed the insights accordingly.

psinger (approved):
Thanks, let's stick to it then!

Just one thing, if somehow possible to fix:
image

In this example, I cant even scroll the full way. Hover works, so I assume thats the workaround.

Speaking of hover, is it possible to enable hover always, even if it fits?

Issue #100, opened 2023-05-08
[Feature] Allow fine tuning T5 models?


Taytay:
### 🚀 Feature

I'd like to experiment with fine tuning small T5 based models, but it looks like there are some assumptions made in the code about the type of model and configuration class required. I haven't investigated yet, so I don't know how big of an ask this is.

### Motivation

I'm training one-off models for specific tasks rather than instruction or chat-tuned models, and think the T5 family of models would be faster and more appropriate. But I like the boilerplate and config in LLM studio.

shivance:
Heyy @psinger I would like to take a crack at this !

shivance:
Hi @psinger , I figured out that we can add models by adding model preset here :

https://github.com/h2oai/h2o-llmstudio/blob/be2f2b155dd79ab6fd1f9fd85c5b410c58b33475/llm_studio/python_configs/text_causal_language_modeling_config.py#L492-L499

and then LLM studio loads the backbone with `AutoModel.from_pretrained(<preset-name>)`. Hence 

> I added `t5-small` in this `ConfigProblemBase` class and ran llm studio. 
> I created an experiment with default dataset and ran experiment
> following error was thrown


<details> <summary>Error </summary>
2023-07-04 22:41:46,555 - INFO: Number of observations in validation dataset: 83
Downloading (…)lve/main/config.json: 100%|███████████████████████████████| 1.21k/1.21k [00:00<00:00, 4.44MB/s]
2023-07-04 22:41:47,043 - INFO: Using float16 for backbone
2023-07-04 22:41:47,086 - ERROR: Exception occurred during H2O LLM Studio run:
Traceback (most recent call last):
  File "/home/shivance/Desktop/h2o-llmstudio/train_wave.py", line 106, in <module>
    run(cfg=cfg)
  File "/home/shivance/Desktop/h2o-llmstudio/train.py", line 594, in run
    model = cfg.architecture.model_class(cfg)
  File "/home/shivance/Desktop/h2o-llmstudio/llm_studio/src/models/text_causal_language_modeling_model.py", line 141, in __init__
    self.backbone, self.backbone_config = create_nlp_backbone(
  File "/home/shivance/Desktop/h2o-llmstudio/llm_studio/src/utils/modeling_utils.py", line 526, in create_nlp_backbone
    backbone = model_class.from_pretrained(
[...]

PR #308, opened 2023-07-24
Sequence to Sequence Problem Type

psinger:
Closes https://github.com/h2oai/h2o-llmstudio/issues/100

Adds a new problem Type Sequence to Sequence Modeling. For now tested for different T5 models.
We can probably go step by step adding more functionality if this will be frequently used.

I tried to reuse as much as possible to not have much duplicate code.

@maxjeblick please take an initial look, it might need 1-2 iterations

Todo:

- [x] Rework dataset?
- [x] Check padding
- [x] Check merging / pipeline
- [x] Default settings for dtype etc?
- [x] Model card


maxjeblick (commented):
Thanks a lot for adding this problem type, it looks in a very good state!

I would move `prepare_lora` and `generate` to separate functions to avoid code duplication (also, they will probably be needed for other potential problem types as well).
Apart from that, the code looks fine, will test it more the next days.

psinger:
Thx for the comments @maxjeblick - tried to address them.

llm_studio/src/utils/config_utils.py
```diff
@@ -204,6 +204,10 @@ def load_config_yaml(path: str):
         from llm_studio.python_configs.text_causal_language_modeling_config import (
             ConfigProblemBase,
         )
+    elif cfg_dict["problem_type"] == "text_sequence_to_sequence_modeling":
```
maxjeblick, line 207:
```
def load_config_yaml(path: str):
    """Loads config from yaml file

    Args:
        path: path of file to load from
    Returns:
        config object
    """
    with open(path, "r") as fp:
        cfg_dict = yaml.load(fp, Loader=yaml.FullLoader)

    problem_type = cfg_dict["problem_type"]
    module_name = f"llm_studio.python_configs.{problem_type}_config"
    try:
        module = importlib.import_module(module_name)
    except ModuleNotFoundError:
        raise NotImplementedError(f"Problem Type {problem_type} not implemented")
    return module.ConfigProblemBase.from_dict(cfg_dict)

```

This code snippet should work, may useful to switch to easy subsequent adding of new problem types.

maxjeblick (changes requested):
Thanks for the changes! Some observations:

- Summary tab fails with `AttributeError: 'ConfigProblemBase' object has no attribute 'hf'`
- Train Data Insights currently shows decoded input ids, whereas the model uses prompt input ids. Maybe it is useful in general to show prompt input ids, instead of input ids? Could potentially also changed in #301.
- Mid-term, promoting `generate` and `prepare_lora` to dedicated functions (with corresponding arguments, `cfg`, `backbone`, ...) may be useful. We can leave it here as is and I can have a look in #317.

llm_studio/src/utils/config_utils.py
maxjeblick: You also leave as is, as it is already part of RLHF PR

maxjeblick:
The PR is probably good to merge after hf has been added to the config.

psinger:
Yes, I just noticed that I also need to adjust the summary model card now. On it.

maxjeblick (approved):
LGTM thanks!

One small issue:
`trust_remote_code` is not populated in the model card, should be easy to fix.

PR #316, opened 2023-07-26
New Experiment Summary page

fatihozturkh2o:
This PR introduces new Experiment Summary cards
- New Summary page includes stats cards to highlight experiments’ main configurations and scores.
- The cards are grouped by the titles of Experiment, Datasets, Score, and Main Configurations.
    - Thought it would be useful to highlight the most important settings of an experiment.
    - (We can expand or replace some of them. Please let me know in the comments) 
- Now there is also a code card that shows how to interact with the experiment model via the HF transformers package.
    - If a model is not pushed to an HF repo yet, the “**model**” attribute in the code pipeline will be shown as “**account/model**”
    - Otherwise, **the actual HF model path** will be put in place.
    - To store HF repo info after pushing models, introduced a new config subclass: **ConfigNLPCausalLMHF**
    - (We might also consider displaying the code card only when the model is pushed to HF, wonder what you’d prefer?)

## How it looks
image

Hope you find it useful and please feel free to let me know if there is anything you'd like to change/add to the PR 🙌

### DIFF
--- a/app_utils/sections/experiment.py
+++ b/app_utils/sections/experiment.py
@@ -1062,18 +1062,6 @@ async def summary_tab(experiment_id, q):
                 justify="between",
                 inset=True,
             ),
-            ui.stats(
-                [
-                    ui.stat(
-                        value="-"
-                        if cfg._parent_experiment == ""
-                        else cfg._parent_experiment,
-                        label="Parent Experiment Name",
-                    ),
-                ],
-                justify="between",
-                inset=True,
-            ),
         ],
     )
     q.client.delete_cards.add(card_name)
@@ -1191,6 +1179,52 @@ async def summary_tab(experiment_id, q):
                 inset=True,
             ),
         ],
+    )
+    q.client.delete_cards.add(card_name)
+
+    # todo: shall we really have this card? If so, we can read it from model card markdown file?
+    # code card
+    device_map = '{"": "cuda:0"}'
+    content = f"""
+To use the model with the transformers library on a machine with GPUs, first make sure you have the transformers, accelerate, torch and einops libraries installed.
+```python\n
+pip install transformers==4.29.2
+pip install accelerate==0.19.0
+pip install torch==2.0.0
+pip install einops==0.6.1
+```
+
+```python\n
+import torch
+from transformers import pipeline
+
+generate_text = pipeline(
+    model="h2oai/h2ogpt-gm-oasst1-en-1024-open-llama-7b-preview-400bt",
+    torch_dtype=torch.{cfg.architecture.backbone_dtype},
+    trust_remote_code=True,
+    use_fast=False,
+    device_map={device_map},
+)
+
+res = generate_text(
+    "Why is drinking water so healthy?",
+    min_new_tokens=2,
+    max_new_tokens=512,
+    do_sample=False,
+    num_beams=1,
+    temperature=float(0.3),
+    repetition_penalty=float(1.2),
+    renormalize_logits=True
+)
+print(res[0]["generated_text"])
+
+```
+"""
+    card_name = "experiment/display/summary/code"
+    q.page[card_name] = ui.markdown_card(
+        box=ui.box(zone="third"),
+        title="",
+        content=content,
     )
     q.client.delete_cards.add(card_name)
 
--- a/app_utils/sections/experiment.py
+++ b/app_utils/sections/experiment.py
@@ -1844,6 +1844,7 @@ def get_experiment_summary_code_card(cfg) -> str:
     # Configs
     text = text.replace("{{min_new_tokens}}", str(cfg.prediction.min_length_inference))
     text = text.replace("{{max_new_tokens}}", str(cfg.prediction.max_length_inference))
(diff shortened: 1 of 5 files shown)


maxjeblick (commented):
Thanks a lot for the PR, really like the new summary tab!

Regarding the structure of the code, I would store the huggingface repo id in the database rather than in the model configuration. This would probably require the addition of a database migration setup, such as alembic (see e.g. in [optuna](https://github.com/optuna/optuna/tree/master/optuna/storages/_rdb)). Happy to discuss this!

app_utils/sections/experiment.py
```diff
@@ -33,16 +34,17 @@
...
     get_unique_name,
-    make_label,
     parse_ui_elements,
     remove_model_type,
     set_env,
     start_experiment,
 )
 from app_utils.wave_utils import busy_dialog, ui_table_from_df, wave_theme
+from llm_studio.python_configs.text_causal_language_modeling_config import (
+    ConfigProblemBase,
```
maxjeblick, line 44: Can you remove this import? We will probably add new problem types in the future (see also PRs).

fatihozturkh2o: Ok, but how to type annotate then for the new function I put? Shall I just remove the annotation?

fatihozturkh2o:
> Thanks a lot for the PR, really like the new summary tab!
> 
> Regarding the structure of the code, I would store the huggingface repo id in the database rather than in the model configuration. This would probably require the addition of a database migration setup, such as alembic (see e.g. in [optuna](https://github.com/optuna/optuna/tree/master/optuna/storages/_rdb)). Happy to discuss this!

Happy to hear you like it!

Yeah, storing it in db sounds better indeed, but have no experience with this db migration setup thing. It'd be nice if we can discuss it 👍

app_utils/sections/experiment.py
maxjeblick: For now, we can remove the type annotation. In general, having a `ConfigProblemBase` base class that each problem type subclasses  may be useful.

maxjeblick:
> Yeah, storing it in db sounds better indeed, but have no experience with this db migration setup thing. It'd be nice if we can discuss it

Yeah happy to discuss it. Haven't used alembic myself, but it looks as if one could use optuna's implementation (in particular the [_VersionManager class](https://github.com/optuna/optuna/blob/master/optuna/storages/_rdb/storage.py#L1023)) as a starting point. I can also take a look into this, if you want.

app_utils/sections/experiment.py
fatihozturkh2o: done https://github.com/h2oai/h2o-llmstudio/pull/316/commits/49d422791fe970426297e600acd04e9745d1f440

fatihozturkh2o:
> > Yeah, storing it in db sounds better indeed, but have no experience with this db migration setup thing. It'd be nice if we can discuss it
> 
> Yeah happy to discuss it. Haven't used alembic myself, but it looks as if one could use optuna's implementation (in particular the [_VersionManager class](https://github.com/optuna/optuna/blob/master/optuna/storages/_rdb/storage.py#L1023)) as a starting point. I can also take a look into this, if you want.

After discussing this idea over a call, we decided to keep the current as is (in terms of not touching db stuff) and work on DB migration functionality in another PR. After having functionality, we can move the HF repo id in DB as a new field.

maxjeblick (changes requested):
Some minor things, apart from that LGTM:

- `use_fast` placeholder is not populated in the model card
- Import torch is actually not used (this is also in the other model card, could you remove it there, as well)?
- Need to use `use_auth_token=True` as model is private by default. A code comment that explains that one needs to log into hugginface hub before probably also makes sense. (See docstring of Hugginface
```
The token to use as HTTP bearer authorization for remote files. If `True`, or not specified, will use
                the token generated when running `huggingface-cli login` (stored in `~/.huggingface`).
```
). If you like, you can also add it to the other model card.


 Screenshot 2023-07-27 at 15-23-02 H2O LLM Studio

fatihozturkh2o:
> Import torch is actually not used (this is also in the other model card, could you remove it there, as well)?

Good catch!

It seems the "einops" library is also not used, shall I remove that too?

fatihozturkh2o:
> Some minor things, apart from that LGTM:
> 
> use_fast placeholder is not populated in the model card
> Import torch is actually not used (this is also in the other model card, could you remove it there, as well)?
> Need to use use_auth_token=True as model is private by default. A code comment that explains that one needs to log into hugginface hub before probably also makes sense. (See docstring of Hugginface
> The token to use as HTTP bearer authorization for remote files. If `True`, or not specified, will use
>                 the token generated when running `huggingface-cli login` (stored in `~/.huggingface`).
> ). If you like, you can also add it to the other model card.
> 

done: https://github.com/h2oai/h2o-llmstudio/pull/316/commits/9d7cfadf3b6b991fe9c0ac1c04481d3e4ed8c76e

- please let me know wdyt about the way I mentioned "token" issue
- stopped mentioning all the unused packages in both MD files.

maxjeblick (approved):
LGTM, thanks!

PR #328, opened 2023-08-01
Refactor rlhf

### LINKED ISSUE #317: [CODE IMPROVEMENT] Promotoe RLHF as a separate problem type

Add a new problem type for RLHF.

### Motivation

- RLHF gets more visible (also in experiment list)
- CausalLM problem type configuration yaml does not contain RLHF settings
- Possibility to hide configuration items (such as enable lora) that are not configurable in RLHF
- General code refactoring

maxjeblick:
This PR adds a separate problem type for RLHF.

Some discussion items:
- How to structure the training settings configuration. The current order may not be optimal.
- `train.py` contains two training functions (`run_train_rlhf` and `run_train`) with partially duplicated code. IMO, it is ok to keep it as is rather than having one function with multiple if-else statements.
- Train data insights should be redone (`Target Text` is not used, thus always an empty string). Will implement it here after #301 has been merged.

Fixes #317
We can first merge #308 and I will fix potential merge conflicts subsequently.


maxjeblick:
Should be good for a first review.


Some points I noticed:
- Train data insights do not display the answer text. We could either leave it as is (highlighting that training is not using ground truth answer), or removing the field by subclassing the plot class.
- Some inference parameters for RLHF are hidden in the chat window. One could explicitly change the visibility in the chat tab?

llm_studio/python_configs/text_rlhf_language_modeling_config.py
```diff
@@ -0,0 +1,197 @@
...
+
+@dataclass
+class ConfigRLHFLMPrediction(ConfigNLPCausalLMPrediction):
+    metric_class: Any = text_causal_language_modeling_metrics.Metrics
+    metric: str = "GPT"
+    metric_gpt_model: str = "gpt-3.5-turbo-0301"
+
+    min_length_inference: int = 2
+    max_length_inference: int = 256
+    batch_size_inference: int = 0
```
pascal-pfeiffer, line 135: aren't these already inherited?

llm_studio/python_configs/text_rlhf_language_modeling_config.py
```diff
@@ -0,0 +1,197 @@
...
+
+    do_sample: bool = True
+    num_beams: int = 1
```
pascal-pfeiffer, line 138: also inherited?

llm_studio/python_configs/text_rlhf_language_modeling_config.py
```diff
@@ -0,0 +1,197 @@
...
+    temperature: float = 0.3
+    repetition_penalty: float = 1.0
+    stop_tokens: str = ""
+    top_k: int = 0
+    top_p: float = 1.0
+
+    num_history: int = 4
```
pascal-pfeiffer, line 145: this too

llm_studio/src/datasets/text_rlhf_modeling_ds.py
```diff
@@ -0,0 +1,67 @@
...
+from llm_studio.src.datasets.text_utils import get_texts
+
+logger = logging.getLogger(__name__)
+
+
+class CustomDataset(CausalLMCustomDataset):
+    def __init__(self, df: pd.DataFrame, cfg: Any, mode: str = "train"):
+        assert (
+            cfg.dataset.system_column == "None"
+        ), "RLHF is not compatible with system column."
```
pascal-pfeiffer, line 20: This is for CLI usage? Shall we just ignore it, or would you still throw the error?

We should also open a new issue to allow system cols with RLHF. The issue is the reward model, so probably depends on the custom reward model training first.

llm_studio/src/datasets/text_rlhf_modeling_ds.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def __getitem__(self, idx: int) -> Dict:
+        """Reads a single text observation."""
+        sample = super().__getitem__(idx)
+        idx = self.indices[idx]
+        sample["reward_model_prompt_text"] = self.get_reward_model_prompt_text(idx)
+        return sample
+
+    def get_labels(self, encodings):
+        if self.mode != "train":  # no labels required for RLHF during training
```
pascal-pfeiffer, line 32: shouldn't that be `==` then?

pascal-pfeiffer (changes requested):
Thanks a lot @maxjeblick . 
I really like your refactors and it makes great sense to put RLHF in an own problem type. 

I will still need to go through the model and train changes and also do some local testing, but will already provide you some initial thoughts now.

llm_studio/src/datasets/text_rlhf_modeling_ds.py
maxjeblick: Good catch. One could also always return labels, as we do not (yet?) support RLHF on prompt dataset, only.

llm_studio/python_configs/text_causal_language_modeling_config.py
```diff
@@ -185,21 +169,13 @@ def __post_init__(self):
         self._possible_values[
             "differential_learning_rate_layers"
         ] = possible_values.String(
-            values=("backbone", "embed", "value_head"),
+            values=("backbone", "embed"),
             allow_custom=False,
             placeholder="Select optional layers...",
```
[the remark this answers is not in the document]
maxjeblick: Will show up on the experiment page
Screenshot 2023-08-08 at 10-30-17 H2O LLM Studio

llm_studio/python_configs/text_rlhf_language_modeling_config.py
```diff
@@ -0,0 +1,197 @@
...
+
+@dataclass
+class ConfigRLHFLMDataset(ConfigNLPCausalLMDataset):
+    dataset_class: Any = CustomDataset
+
+    def __post_init__(self):
+        super().__post_init__()
+        # RLHF is not compatible with system column.
+        self.system_column = "None"
+        self._visibility["system_column"] = False
```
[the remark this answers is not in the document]
maxjeblick: done

llm_studio/python_configs/text_causal_language_modeling_config.py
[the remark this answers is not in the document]
pascal-pfeiffer: right
Only thinking about the wording then, but fine for this PR for sure.

llm_studio/src/datasets/text_rlhf_modeling_ds.py
maxjeblick: Yeah, it is basically an additional safety check.

pascal-pfeiffer (commented):
image
- LLM Backbone is not a dropdown for me

- hide output dir

We could probably move the reward model to the top, too, wdyt?

maxjeblick:
Thanks for the review, it was very helpful. I addressed the issues above, in addition I added some smaller code changes:

- I fixed a small bug in `plot_batch`
- I refactored `.from_dict` class method to a common base method. 

> We could probably move the reward model to the top, too, wdyt?

Yes good idea, I changed that.

pascal-pfeiffer (approved):
Thanks a lot @maxjeblick , very nice refactor and small fixes along the way!
Looks all good to me now. 

I will likely work on RLHF again in a follow up PR/issue and allow for larger batches. It should be much easier now with the individual train loops.

PR #342, opened 2023-08-04
saving hf yaml separately

fatihozturkh2o:
With this PR:
- Deleted ConfigNLPCausalLMHF class and stopped storing it within ConfigProblemBase objects (it was not part of the actual experiment settings anyway, so definitely cleaner now )
- Saving & loading hf.yaml files per experiment to store hf repo information (in case a model is pushed to the HF)
- Path to each hf.yaml is "{cfg.output_directory}/hf.yaml"
- Fixed a bug where the experiment config object was being re-saved with incorrect configs after each hf-model-push.


psinger:
I merged my PR before, there were some changes to this. Need to merge here and address.

psinger (changes requested):
There are multiple cfg files now for different problem types, needs to be addressed in all.

fatihozturkh2o:
> There are multiple cfg files now for different problem types, needs to be addressed in all.

Can you elaborate more? With some examples if possible? 

My current changes are not based on any cfg right now, just storing a separate hf.yaml file. Can you point me to what's missing?

pascal-pfeiffer:
> > There are multiple cfg files now for different problem types, needs to be addressed in all.
> 
> Can you elaborate more? With some examples if possible?
> 
> My current changes are not based on any cfg right now, just storing a separate hf.yaml file. Can you point me to what's missing?

You need to apply your changes not only to `llm_studio/python_configs/text_causal_language_modeling_config.py` but also in the new sequence to sequence config file.

fatihozturkh2o:
> > > There are multiple cfg files now for different problem types, needs to be addressed in all.
> > 
> > 
> > Can you elaborate more? With some examples if possible?
> > My current changes are not based on any cfg right now, just storing a separate hf.yaml file. Can you point me to what's missing?
> 
> You need to apply your changes not only to `llm_studio/python_configs/text_causal_language_modeling_config.py` but also in the new sequence to sequence config file and maybe the base file
> 
> `llm_studio/python_configs/text_sequence_to_sequence_modeling_config.py` `llm_studio/python_configs/base.py`

I don't have any changes in text_sequence_to_sequence_modeling_config.py though. I previously had, but removed them in this PR. Which line exactly do I need to apply to all? 🤔

psinger:
Yes, and you also need to remove it in the other cfg file if you want to remove `hf`.

pascal-pfeiffer:
> > > > There are multiple cfg files now for different problem types, needs to be addressed in all.
> > > 
> > > 
> > > Can you elaborate more? With some examples if possible?
> > > My current changes are not based on any cfg right now, just storing a separate hf.yaml file. Can you point me to what's missing?
> > 
> > 
> > You need to apply your changes not only to `llm_studio/python_configs/text_causal_language_modeling_config.py` but also in the new sequence to sequence config file and maybe the base file
> > `llm_studio/python_configs/text_sequence_to_sequence_modeling_config.py` `llm_studio/python_configs/base.py`
> 
> I don't have any changes in text_sequence_to_sequence_modeling_config.py though. I previously had, but removed them in this PR. Which line exactly do I need to apply to all? 🤔

https://github.com/h2oai/h2o-llmstudio/blob/main/llm_studio/python_configs/text_sequence_to_sequence_modeling_config.py#L106

for example this line probably?

fatihozturkh2o:
@psinger @pascal-pfeiffer Ok I see, now. The confusion happened because I never touched seq-to-seq config file before (because it didn't exist when I introduced hf stuff) and changes came with the master rebasing.

Thanks for pointing out clearly.

done: https://github.com/h2oai/h2o-llmstudio/pull/342/commits/c097a86cd46c9dbc9f5dc41e75cbf9ef28eaf680

psinger (approved):
lgtm! thx

PR #351, opened 2023-08-08
Remove pickle support

### LINKED ISSUE #350: [CODE IMPROVEMENT] Deprecate pickle migration

Deprecate `pickle -> yaml` conversion script during startup.

### Motivation

- It slows down the app startup
- Migration should not be needed any longer/should already happened
- Old pickle format may be broken anyhow as code got refactored a couple of times now

maxjeblick:
Fixes #350


pascal-pfeiffer (approved):
good idea, lgtm

PR #355, opened 2023-08-09
Fix tests and test workflow

### LINKED ISSUE #354: [BUG] GitHub test workflow does not report errors
### 🐛 Bug

Due to refactoring, some tests are currently broken. GitHub workflow does not report any issues, even if the tests are failing.

### To Reproduce

See e.g. this PR: https://github.com/h2oai/h2o-llmstudio/pull/300/checks

maxjeblick:
This PR fixes some tests and ensures that github test workflow fails correctly.

Fixes #354


Makefile
```diff
@@ -71,7 +71,7 @@ black: pipenv
 test: reports
 	export PYTHONPATH=$(PWD) && $(PIPENV) run pytest -v -s -x \
 		--junitxml=./reports/junit.xml \
-		tests/* | tee reports/pytest.log
+		tests/* || exit 1
```
pascal-pfeiffer, line 74: as we use the same tests locally, I may want to have these reports still. 
Should we use something like we use for mypy testing:

https://github.com/h2oai/h2o-llmstudio/blob/main/Makefile#L54-L55

pascal-pfeiffer (approved):
Thanks a lot, good catch!

PR #369, opened 2023-08-16
Pp/rlhf large rollout

pascal-pfeiffer:
incorporating a few updates from upstream trl and allowing longer rollouts

Now, you can specify the `rollout_steps` (using batch_size for each) which should be done before each PPO cycle (using ppo_batch_size)

Combined with grad_accum and the newly introduced parameter `full_kl_penalty`, this should lead to much more stable training.


llm_studio/python_configs/text_causal_language_modeling_config.py
```diff
@@ -323,7 +323,6 @@ def __post_init__(self):
         self._possible_values["repetition_penalty"] = (1, 10, 0.05)
         self._possible_values["top_k"] = (0, 100, 1)
         self._possible_values["top_p"] = (0.5, 1, 0.05)
-        self._possible_values["num_history"] = (1, 50, 1)
```
maxjeblick, line 326: Chat mode fails with
```
File “/data/maxjeblick/PyCharmProjects/h2o-llmstudio/./llm_studio/app_utils/handlers.py”, line 321, in handle await experiment_display(q)

File “/data/maxjeblick/PyCharmProjects/h2o-llmstudio/./llm_studio/app_utils/sections/experiment.py”, line 983, in experiment_display await chat_tab(q)

File “/data/maxjeblick/PyCharmProjects/h2o-llmstudio/./llm_studio/app_utils/sections/chat.py”, line 140, in chat_tab option_items = get_ui_elements(

File “/data/maxjeblick/PyCharmProjects/h2o-llmstudio/./llm_studio/app_utils/utils.py”, line 1180, in get_ui_elements t = get_ui_element(

File “/data/maxjeblick/PyCharmProjects/h2o-llmstudio/./llm_studio/app_utils/utils.py”, line 798, in get_ui_element raise ValueError(

ValueError: Type annotations int and float need a possible_values.Number!
```

llm_studio/src/trl/trainer.py
```diff
@@ -651,15 +682,62 @@ def compute_rewards(
...
 
+    def _kl_penalty(
+        self, logprob: torch.FloatTensor, ref_logprob: torch.FloatTensor
+    ) -> torch.FloatTensor:
+        if self.cfg.training.full_kl_penalty:
+            # Flip is required due to this issue?
+            # https://github.com/pytorch/pytorch/issues/57459
+            return F.kl_div(
+                ref_logprob, logprob, log_target=True, reduction="none"
+            ).sum(-1)
```
maxjeblick, line 693: Why sum and not mean here?

maxjeblick: I've another question `ref_logprob, logprob` order. Why is it this order i.e. 
```
input = ref_logprob
target = logprob
```

pascal-pfeiffer: Good question about the sum. I took it from the trl implementation. I'd need to check, but this should not be the token dimension that is being reduced. 

https://github.com/pytorch/pytorch/issues/57459 it seems to be inverting the positional arguments. Apparently this way is correct, as the KL_div is positive.

maxjeblick: Thanks, I think it should be good then

maxjeblick (approved):
LGTM, thanks a lot!

PR #373, opened 2023-08-16
ignore warnings for loading config yaml

### LINKED ISSUE #372: [CODE IMPROVEMENT] Reduce too verbose logging on startup

Reduce logging on startup in case experiment config yaml has keys that do not match the current dataclass.

When starting the app, `get_experiments(q)` goes trough all experiments and loads the config yaml.
In case experiment configuration dataclasses code has been changed (as it has been the case recently), the startup logs can be polluted with

```
2023-08-16 17:56:12,061 - WARNING: Keys {'reward_model', 'ppo_clip_policy', 'advantages_gamma', 'ppo_epochs', 'kl_target', 'use_rlhf', 'initial_kl_coefficient', 'ppo_batch_size', 'adaptive_kl_control', 'scaling_factor_value_loss', 'offload_reward_model', 'advantages_lambda', 'ppo_generate_temperature', 'kl_horizon', 'ppo_clip_value'} are not in the config.
2023-08-16 17:56:12,061 - WARNING: Keys {'number_of_texts'} are not in the config.
```

### Motivation
Remove the startup warnings as
- It is unclear where these warning exactly comes from and what to do
- Warning should only be present when a user starts an experiment from a configuration (or old experiment) that has unseen keys.

maxjeblick:
Catch `"*are not in the config."`warnings that may arise when loading old configurations  in `get_datasets_info` and `get_experiments_info` functions.

Fixes #372

### DIFF
--- a/app_utils/utils.py
+++ b/app_utils/utils.py
@@ -14,6 +14,7 @@ import socket
 import subprocess
 import time
 import uuid
+import warnings
 import zipfile
 from collections import defaultdict
 from contextlib import closing
@@ -1288,7 +1289,12 @@ def get_experiments_info(df: DataFrame, q: Q) -> DefaultDict:
     info = defaultdict(list)
     for _, row in df.iterrows():
         try:
-            cfg = load_config_yaml(f"{row.path}/cfg.yaml").__dict__
+            # load_config_yaml issues a warning if the yaml file contains keys
+            # that are no longer part of the dataclass fields.
+            # This can happen if the codebase has changed since the experiment was run.
+            # Ignore those warnings here
+            with warnings.filterwarnings("ignore", message="*are not in the config."):
+                cfg = load_config_yaml(f"{row.path}/cfg.yaml").__dict__
         except Exception:
             cfg = None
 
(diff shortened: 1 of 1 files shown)


app_utils/utils.py
fatihozturkh2o, line 1: We also use similar log filtering in MV. The only thing is, I'd suggest having a variable for this either from the app.toml or wherever you store app configs/ env variables. 

Maybe we can have a variable like `LOG_NOT_IN_CONFIG_WARNING = False` and make use of it here. There might be some issues in the future where we'd like to see the logs actually, and it'd be easy to overlook.

Wdyt?

maxjeblick: Thanks for the feedback! Here, I don't think this fine-grained control over the logs is required as
the functions `get_datasets_info` and `get_experiments_info` should not issue the filtered warnings (they only 
use some sparse information of the cfg file, maybe that information should also be part of the database).

The alternative would be to rewrite/extract the filtering of the yaml config and use that function explicitly. 

Also happy  to discuss on a call with the respective code in MV.

fatihozturkh2o: Or maybe logs should be directly handled in load_config_yaml function. I see your point about these functions shouldn't be responsible for this control. We can discuss this later 👍

fatihozturkh2o (approved):
lgtm!

pascal-pfeiffer (approved):
lgtm, too

PR #364, opened 2023-08-10
Enhancements to Secret Management

maxjeblick:
This pull request adds different backend handlers for storing passwords with `keyring` being the default.
On systems where keyring fails, it is possible to select a local .env backend, or not to save the tokens at all.

**Purpose**
The purpose of these changes is to provide the application with greater flexibility and security when it comes to managing and storing sensitive information.

**Backward compatibility**
A migration process is implemented that converts the old pickle format.
On systems where migration to keyring fails, the user needs to manually reenter missing tokens (this is mentioned in the logs).

Fixes #363

PR #380, opened 2023-08-18
add integration tests for all problem types

pascal-pfeiffer:
run a small experiment


tests/integration/test_integration.py
```diff
@@ -0,0 +1,95 @@
...
+    ],
+)
+def test_oasst_training(tmp_path, config_name):
+    """
+    Test training on OASST dataset.
+
+    Pytest keeps around the last 3 test runs in the tmp_path fixture.
+    """
+
+    if torch.cuda.is_available():
```
maxjeblick, line 50: One could use 
`need_gpus = pytest.mark.skipif(not torch.cuda.is_available(), reason="GPU only test")`
`@need_gpus` here. Test would then be skipped completely.

pascal-pfeiffer: til :) 
thank you

app.py
```diff
@@ -1,16 +1,16 @@
 import logging
 import os
 
-from app_utils.sections.chat import show_chat_is_running_dialog
+from llm_studio.app_utils.sections.chat import show_chat_is_running_dialog
```
maxjeblick, line 4: Like the idea to move `app_utils` to llm_studio! Is this required for the tests?
Also, in a recent discussion with @fatihozturkh2o, we where thinking of removing the utils part from `app_utils` folder or renaming it to something like `frontend`. WDYT?

pascal-pfeiffer: it's required for getting the test coverage right. I would also add this to the command. Currently looking into best ways for the report.

fatihozturkh2o: yeah, this is what we planned pretty much:

https://github.com/h2oai/h2o-llmstudio/pull/364#discussion_r1295726033

pascal-pfeiffer: we also need to move the train.py and train_wave.py to the code subfolder for discoverability in pytest-coverage.
I'll push this and add symlinks probably to keep the old behavior

pascal-pfeiffer: moving `train.py` and the other python files in the root dir is a bit more tricky than expected. I'd probably postpone for later, as it really only affects the coverage report.

psinger (approved):

PR #386, opened 2023-08-18
conversation chaining refactoring

### LINKED ISSUE #370: [CODE IMPROVEMENT] Refactor chained samples

Extract logic for computing conversation chains into  separate functions/helper class.

### Motivation

Currently, it is not possible to easily extract complete conversations without initializing the `CustomDataset` class.
The dataset class introduces some side effects (e.g. it automatically adds eos token to system text) and only shows complete samples if `cfg.dataset.limit_chained_samples` is True.

Useful for subsequent tasks such as:
- #337
- #229
- Also better to test in isolation

maxjeblick:
This PR refactors the logic for conversation chaining into a separate class.

In addition, the dataset class is slightly refactored:
- `init` does not check for missing values etc., this is moved to sanity check method
- `encodings` list is split into `prompt_encodings` and `answer_encodings` for better readability
- `get_input_columns` was removed to make the class slimmer.

I have not changed any of the logic to include Llama2-chat prompting style. Maybe we can create a separate PR for this?

Fixes #370

### DIFF
--- a/app_utils/sections/chat.py
+++ b/app_utils/sections/chat.py
@@ -239,7 +239,7 @@ async def chat_update(q: Q) -> None:
 
     def text_cleaner(text: str) -> str:
         return cfg.dataset.dataset_class.clean_output(
-            output={"predicted_text": np.array([text])}, prompts=[full_prompt], cfg=cfg
+            output={"predicted_text": np.array([text])}, cfg=cfg
         )["predicted_text"][0]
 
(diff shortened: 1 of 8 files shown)


maxjeblick:
Sorry (again) for the late reply and thanks for the catch!
The issue was that the default dataset does contain parent_ids that are not in the dataset itself. I fixed this issue and added a test for the OASST dataset.

Issue #374, opened 2023-08-17
[FEATURE] Allow more detailed control over quantization


pascal-pfeiffer:
### 🚀 Feature

We currently have a few settings for 4bit and 8bit quantization hard-coded 

```python
            load_in_8bit=True,
            llm_int8_threshold=0.0,
```

```python
            load_in_4bit=True,
            bnb_4bit_compute_dtype=torch.float16,
            bnb_4bit_quant_type="nf4",
```

We may want to open these as nested settings and also allow `bnb_4bit_use_double_quant` for further reduction of the memory footprint.

### Motivation

More control over quantization.

binga:
Thanks for opening this issue. Does the addition of DQ to 4-bit finetuning need additional testing or just a PR?

pascal-pfeiffer:
> Thanks for opening this issue. Does the addition of DQ to 4-bit finetuning need additional testing or just a PR?

It should be relatively stable. At least, I haven't found any issues with it so far. So, just (?) the PR.

binga:
As NF4+DQ produces better results (in terms of memory and performance), to keep the interface relatively beginner-friendly, I've added DQ as a default, when int4 finetuning is selected, in this PR - https://github.com/h2oai/h2o-llmstudio/pull/403.

IMO, more control over all the levers should be an advanced setting - where there are front-end design changes and corresponding backend changes. Is this reasonable?

PR #403, opened 2023-09-04
Added double quantization by default for int4 finetuning

binga:
This PR adds Double Quantization as a default to Int4 finetuning based on results in Table 3 in https://arxiv.org/pdf/2305.14314.pdf.


psinger:
I dont see a comparison with not using DQ in that paper. Personally I would leave it at `False` by default.

binga:
Any specific reason why DQ is being defaulted to `False`?

psinger:
Because it is an extra setting that could potentially have side effects.
It is also False by default in HF
https://github.com/huggingface/transformers/blob/v4.32.1/src/transformers/utils/quantization_config.py#L173

We will add settings to enable it via https://github.com/h2oai/h2o-llmstudio/issues/374

binga:
Understood, thank you for sharing additional context. Closing the PR.

PR #410, opened 2023-09-08
fix perplexity metric in RLHF

### LINKED ISSUE #384: [BUG] RLHF fail with metric set to Perplexity
### 🐛 Bug

fails with perplexity in rlhf

pascal-pfeiffer:
we still need the predicted answer for the reward model during training force that with force_transform=True in the train loop

fixes #384


llm_studio/src/datasets/text_rlhf_modeling_ds.py
```diff
@@ -55,13 +55,19 @@ def get_reward_model_prompt_text(self, idx):
...
-        if cfg.prediction.metric == "Perplexity":
+    def postprocess_batch_predictions(
+        self, cfg: Any, output: Dict, force_transform: bool = False
+    ) -> Dict:
+        if cfg.prediction.metric == "Perplexity" and not force_transform:
+            # we still need the predicted answer for the reward model during training
+            # force that with force_transform=True in the train loop
             return output
+
         predicted_text = [
```
maxjeblick, line 66: Would it be possible to add an if condition such that predicted text is always computed when `"predicted_answer_ids"` are in the output dictionary? I should not break rlhf training, I guess. By this, one would not need `force_transform`.

pascal-pfeiffer: Certainly possible. Actually I wanted to make it less implicit and more explicit. 

But maybe we can simplify it now that we have RLHF and SFT detangled. Let's me try another solution with always transforming the ids if available.

Needs to work for:
Perplexity - SFT
Perplexity - RLHF
Perplexity - S2S
BLEU - SFT
BLEU - RLHF
BLEU - S2S

I'll expand the integration tests to cover all 6

maxjeblick: Sounds good!

maxjeblick (approved):
lgtm thanks!

Issue #409, opened 2023-09-07
[BUG] Auto layers for LoRA are not working with newest package versions


pascal-pfeiffer:
### 🐛 Bug

7B llama2 model:
```
ValueError: Target modules ['.q_proj', '.k_proj', '.v_proj', '.o_proj', '.gate_proj', '.up_proj', '.down_proj'] not found in the base model. Please check the target modules and try again.
```

related PR: https://github.com/h2oai/h2o-llmstudio/pull/408

pascal-pfeiffer:
only happens with a new `make setup` so it must be a new change in peft @psinger

PR #411, opened 2023-09-08
Packages and Lora fixes

psinger:
Closes https://github.com/h2oai/h2o-llmstudio/issues/409


Pipfile
```diff
@@ -45,11 +45,11 @@ datasets = ">=2.11.0, <3.0.0"
...
-h2o-wave = "0.26"
-tiktoken = "0.4.0"
-hf-transfer = "0.1.3"
-peft = "0.4.0"
-azure-storage-file-datalake = ">=12.12.0"
+h2o-wave = "==0.26.2"
+tiktoken = "==0.4.0"
+hf-transfer = "==0.1.3"
+peft = "==0.5.0"
+azure-storage-file-datalake = "==12.12.0"
```
pascal-pfeiffer, line 52: should be fine to leave as `>=` as this is a mature package which shouldn't change their API often.

pascal-pfeiffer (approved):
lgtm

just think we don't need to fix the azure-storage-file-datalake version

PR #225, opened 2023-07-03
version bump to 0.0.4 and default batch_size 2

pascal-pfeiffer:
version = "0.0.4"
batch_size: int = 2

PR #421, opened 2023-09-20
Improve Data Visualizations

### LINKED ISSUE #229: [FEATURE] Option to view parent prompts on 'Validation Insights' and 'Sample Train Visualization'
### 🚀 Feature

For nested prompts (tree-like), H2O LLM Studio should have an option to display the parent conversations. Currently, on both 'Sample Train Visualization' (datasets tab) and 'Validation Insights' (for a completed experiment) it only shows the 'current' conversation (prompt+answer), without the proper context of previous interactions. For the user of LLM Studio it might be necessary to see such context to evaluate whether the conversation branch makes sense and/or the model gives good answers.

### LINKED ISSUE #337: [CODE IMPROVEMENT] Improve train data statistics

Possible improvements for train data statistics:

- show text length statistics/histograms
- show chained conversation statistics (num of conversations, etc.)
- show language distribution (may be slow/imprecise though)

### Motivation

Current statistics are not that useful

maxjeblick:
This PR improves the data visualizations.

- `Sample Train Visualization` now shows the complete chat history. For each conversation length (== number of prompt-answer pairs per chat), up to 5 samples are shown
- `Train Data Statistics` now shows different histograms for text length distribution (== number of words).
- `Validation Prediction Insights` now show the complete chat history.

Screenshot 2023-09-20 at 11-57-48 H2O LLM Studio
Screenshot 2023-09-20 at 11-58-01 H2O LLM Studio
Screenshot 2023-09-20 at 12-33-28 H2O LLM Studio

Fixes #225
Fixes #229 
Fixes #337


maxjeblick:
Thanks for the detailed feedback, I'll have another look!

> Nevertheless, I experienced quite some loading times for larger datasets (half an hour and still loading for a 1.7 GB dataset).

At which part did you experience the slowdown? I've tested the code with 80k samples, and it worked reasonable fast. I'll create a larger (maybe dummy) dataset and test accordingly.

maxjeblick:
Thanks for the valuable feedback and the dataset to test on.
After #426 and some additional changes, dataset visualizations should now be fast to load.
- Sample train visualization will be loaded from disc if computed once.
- Train data statistics is computed with LRU caching.

For the 1.7GB dataset (or the train df thereof), initial loading of the visualizations takes some time (~1 Minute). Given the size of the DataFrame, this is still a reasonable time to wait IMO. With more effort, one could move the computation to a background task to avoid blocking of the app.

I've also refactored `dataset_display` a bit, assuming that the datasets in LLM Studio will always contain `dataset["train_dataframe"]` (for main branch, data visualizations would break if this was not the case).
Apart from that, parts of the code of `show_statistics_tab` should probably be moved to plot class to unify the codebase. As this would require additional code changes, I did not address it in this PR.

pascal-pfeiffer (changes requested):
Thanks for the quick fixes and improvements @maxjeblick 
The loading time looks reasonable now and I like the caching. I think we could consider using multithreading, if speed really is an issue for larger datasets, but I am fine with the current speed. 

One odd thing that I noticed is the quantile coloring. This doesn't seem to be the 10% quantile.
image


The space taken here still looks a bit odd, too:
image

One more idea for improvement (but probably not worth adding) would be a toggle for a log y-axis.

maxjeblick:
Quantiles should now be fixed.

pascal-pfeiffer:
> Quantiles should now be fixed.

thanks, looks good!

pascal-pfeiffer (approved):

PR #438, opened 2023-10-06
Split Train-val data on parent id

### LINKED ISSUE #422: [CODE IMPROVEMENT] Respect conversation chains for train/valid split

When selecting automatic train/validation split, ensure that complete conversations are either in train or in valid.

### Motivation

Do not break up conversations, as this may 
- Drop parts of the dataset when limit chained samples is True (if an in-between conversation is missing)
- Validation metric may be underestimating model's performance

maxjeblick:
This PR adds automatic train/val splitting based on conversation chain ids (each conversation chain is either in train or in val). 

In addition, the validation insights are slightly reworked such that they visualize the whole conversation history.

Fixes #422


llm_studio/src/plots/text_causal_language_modeling_plots.py
```diff
@@ -176,7 +181,12 @@ def plot_validation_predictions(
             # exclude last answer
             answers[-1] = ""
             for prompt, answer in zip(prompts, answers):
-                input_text += prompt + answer
+                input_text += (
+                    f" **{prompt_column_name}:** "
+                    f"{prompt} "
+                    f"**{cfg.dataset.answer_column}:** "
```
pascal-pfeiffer, line 187: let's add a newline or two here for better readability

pascal-pfeiffer:
I noticed an odd behaviour for the mouse-over tooltips. They sometimes show mixed entries from target and prediction based on what I see in the cell.
maybe, it only happens for code cells.

image
image

llm_studio/src/plots/text_causal_language_modeling_plots.py
maxjeblick: I've added newlines after each prompt/answer.
Screenshot 2023-10-10 at 10-14-55 Screenshot from 2023-10-10 10-13-45 png (PNG Image 5120 × 2880 pixels) — Scaled (46%)

maxjeblick:
> I noticed an odd behaviour for the mouse-over tooltips.

Looks weird indeed. Probably an issue with wave.

tests/utils/test_data_utils.py
```diff
@@ -0,0 +1,111 @@
...
+            for validation_size in [0.01, 0.1, 0.2, 0.3, 0.4, 0.5]:
+                cfg_mock.dataset.validation_size = validation_size
+
+                train_df, val_df = load_train_valid_data(cfg_mock)
+                assert set(train_df["parent_id"].dropna().values).isdisjoint(
+                    set(val_df["id"].dropna().values)
+                )
+                assert set(val_df["parent_id"].dropna().values).isdisjoint(
+                    set(train_df["id"].dropna().values)
+                )
```
pascal-pfeiffer, line 111: Let's actually test for the size of the validation set, too.

something along these lines

```python
assert (len(val_df) / (len(train_df) + len(val_df))) == pytest.approx(
    validation_size, 0.001
)
```

Also, for some reason, this test is not executed properly.

`prepare_default_dataset(tmp_path)` does create a tmp_folder, but it will be empty

`assert len(os.listdir(tmp_path)) > 0` fails. (please add it as a check, too)

One solution is to move it to another test file (but I would still be interested why this is happening.

pascal-pfeiffer (changes requested):
Thank you! everything looks good to me. 
Very clever splits.

I only noticed issues with your test `test_oasst_data_auto_split` that need to be addressed, please.

tests/utils/test_data_utils.py
maxjeblick: Thanks for the feedback, I've added the assert. As for the tmp_path assert, I've also added it to the test to check that it does not fail.

pascal-pfeiffer: Thank you!
I just ran it again with a new env and `make setup-dev` and now the tmp folder is properly populated. I can't explain the old behaviour. Nevertheless, good to have that extra check in place.

Issue #435, opened 2023-10-06
LLM Studio crash every time I try to start training


Ender436:
### 🐛 Bug

I am using settings that are only slightly different than previous successful runs, I'm not quite understanding what is wrong. Any help would be appreciated! Here is the error I'm getting:

stacktrace
Traceback (most recent call last):

File “/home/user/h2o-llmstudio/./llm_studio/app_utils/handlers.py”, line 172, in handle should_list_current_experiments = await experiment_run(

File “/home/user/h2o-llmstudio/./llm_studio/app_utils/sections/experiment.py”, line 536, in experiment_run start_experiment(cfg=cfg, q=q, pre=pre)

File “/home/user/h2o-llmstudio/./llm_studio/app_utils/utils.py”, line 1728, in start_experiment p = start_process(

File “/home/user/h2o-llmstudio/./llm_studio/app_utils/utils.py”, line 145, in start_process p = subprocess.Popen(

File “/usr/lib/python3.10/subprocess.py”, line 971, in init self._execute_child(args, executable, preexec_fn, close_fds,

File “/usr/lib/python3.10/subprocess.py”, line 1783, in _execute_child env_list.append(k + b’=’ + os.fsencode(v))

File “/usr/lib/python3.10/os.py”, line 811, in fsencode filename = fspath(filename) # Does type-checking of filename.

TypeError: expected str, bytes or os.PathLike object, not NoneType

Error
None

### To Reproduce

yaml file is not written, otherwise I would include it

### LLM Studio version

Git Version
9d6e9470447ebbc96e54346365fcb168868a9033

Ender436:
I also seem to be getting errors where it is saying that the directory already exists, but not only did it not exist before I tried running that experiment, it isn't even the name that I had put into the field for naming the experiment. Sorry if I'm missing something obvious.

maxjeblick:
Thanks for posting the issue, I'll investigate it. Did you update llm studio from a previous version or is it a fresh install?

Ender436:
Thanks, I’ll have to try that, it is a fresh install, I installed just yesterday. I also noticed it might be something to do with settings changed in the GUI not being saved or something, because I’ve noticed a few times when creating an experiment and typing in the hugging face model, I’d hit enter or click out of it, and the field would clear as soon as I did so. Thanks again for the help.

Ender436:
I followed the steps you provided, and the error still happened.

maxjeblick:
[...]

PR #440, opened 2023-10-09
Cast missing env variables to String

maxjeblick:
Potential hotfix for #435 

So far, I wasn't able to reproduce the issue. Given the error message `TypeError: expected str, bytes or os.PathLike object, not NoneType`, I suspect `env_vars` contain some `None` variables.

I will investigate the issue further, and potentially check with @RobMulla who also encountered it.


pascal-pfeiffer (approved):
Thank you. Can confirm that it fixes the identified error path.

PR #444, opened 2023-10-12
Enable skipped tests

maxjeblick:
Also fixes path for integration test configs


pascal-pfeiffer (approved):
let's also remove it from `def test_oasst_data_automatic_split(tmp_path):` 

lgtm then

maxjeblick:
> let's also remove it from def test_oasst_data_automatic_split(tmp_path):

Will be done in #438 once tests are fixed

PR #449, opened 2023-10-16
Causal Classification Problem Type

psinger:
New Problem Type for supporting Causal Classification.

Workflow:
- Answer column contains integer encoder binary or multiclass target
- Problem Type is Causal Classification Modeling
- Metrics are AUC and Accuracy
- Number of classes need to be selected via cfg
- Pushing to HF and subsequent loading with additional head supported


llm_studio/app_utils/hugging_face_utils.py
```diff
@@ -122,8 +122,19 @@ def publish_model_to_hugging_face(
         repo_id=repo_id, repo_type="model", commit_message="Upload model card"
     )
 
-    # push config to hub
     api = huggingface_hub.HfApi()
+
+    # push classification head to hub
+    if cfg.type == "causal_classification":
```
maxjeblick, line 128: You could use the already implemented `cfg.problem_type` (that maps to the filename of the config, e.g. text_causal_language_modeling).

llm_studio/src/datasets/text_causal_language_modeling_ds.py
```diff
@@ -31,6 +31,9 @@ def __init__(self, df: pd.DataFrame, cfg: Any, mode: str = "train"):
         self.tokenizer = get_tokenizer(self.cfg)
         self.conversation_chain_handler = ConversationChainHandler(self.df, cfg)
 
+        if cfg.type == "causal_classification":
```
maxjeblick, line 34: We could also think about subclassing the dataset. As the code change is minimal atm, I'm also fine with leaving it at is.

model_cards/text_causal_classification_experiment_summary_card_template.md
```diff
@@ -0,0 +1,61 @@
...
+model = AutoModelForCausalLM.from_pretrained(
+    model_name,
+    torch_dtype="auto",
+    device_map={"": "cuda:0"},
+    trust_remote_code={{trust_remote_code}},
+).cuda().eval()
+
+head_weights = torch.load("classification_head.pth", map_location="cuda")
+# settings can be arbitrary here as we overwrite with saved weights
+head = torch.nn.Linear(1, 1, bias=False).to("cuda")
```
maxjeblick, line 51: `nn.Linear(backbone_config.vocab_size, cfg.dataset.num_classes)`. Easiest option is probably to fetch dimensions from head_weights.shape.

llm_studio/src/losses/text_causal_classification_modeling_losses.py
```diff
@@ -0,0 +1,53 @@
...
+logger = logging.getLogger(__name__)
+
+
+class CrossEntropyLoss(nn.Module):
+    def __init__(self, cfg: Any):
+        super().__init__()
+        self.cfg = cfg
+        self.loss_fn = nn.CrossEntropyLoss()
+
+    def forward(self, logits, labels):
```
maxjeblick, line 18: For sake of code simplicity, I think it's sufficient for now to let the user specify Num Classes manually.

I'd add some exception to the loss, however to inform the user if he/she forgets about this (like I did :) ). Something like
"labels contain [7,5,4], you only selected Num Classes = 1. Please increase Num Classes to account for all classes in the dataset." Otherwise the error reads `RuntimeError: CUDA error: device-side assert triggered` which may confuse the user. This check could also go into the dataset class.

model_cards/text_causal_classification_experiment_summary_card_template.md
psinger: yeah but it is irrelevant, as the weight copy overwrites it anyways

maxjeblick: ah yes, missed the comment

maxjeblick (changes requested):
Thanks a lot for adding the classification problem type! The code looks good (I've left some minor comments), and training also works as expected.

As the setup (loss-metric-num classes) is currently manual, I'd suggest to add some sanity checks (that raise an exception) in the code to check for user input errors. At some later point, we can potentially add automatic handling of the settings, based on the input data. I've ran into the following issues:
- Using a dataset where target values are [1, 2, 3 ,4, 5]. `num_classes` needs to be set to 6, as they are 0-indexed.
- Forgot to change metrics/loss accordingly. For metric, experiment will fail at 100%, only.

Apart from that, we probably want to show the predicted class in the validation prediction insights and also add it to `validation_predictions.csv`.

Some smaller things (can also be addressed later):
- Tokenized input text also shows the classification label.
- We probably want to hide the Chat Window (funny conversation though :D )

psinger:
Thanks @maxjeblick - will address.

I discussed with @pascal-pfeiffer that we can expect here for users to provide the correct inputs, but I agree that exceptions should be more clear.

llm_studio/python_configs/text_causal_classification_modeling_config.py
```diff
@@ -0,0 +1,185 @@
...
+@dataclass
+class ConfigNLPCausalClassificationPrediction(ConfigNLPCausalLMPrediction):
+    metric_class: Any = text_causal_classification_modeling_metrics.Metrics
+    metric: str = "AUC"
+
+    def __post_init__(self):
+        super().__post_init__()
+        self._possible_values["metric"] = self.metric_class.names()
+
+        for k in [
```
maxjeblick, line 110: I'd rather inherit from `DefaultConfig` here (and add
```
    min_length_inference: int = 2
    max_length_inference: int = 256
    batch_size_inference: int = 0
```
) as configuration yaml and config tab will still show non-visible items.

llm_studio/src/losses/text_causal_classification_modeling_losses.py
maxjeblick: fixed by dataset assertions

maxjeblick (approved):
LGTM, thanks a lot for the implementation!

I've left a small comment about `ConfigNLPCausalClassificationPrediction`, apart from that it looks good!

llm_studio/python_configs/text_causal_classification_modeling_config.py
psinger: I think we wont even need `min_length_inference` and `max_length_inference` or did you see it somewhere?

psinger: but probably we need all of them for the model card generation...

psinger: decided to address it in future

samvelkoch:
Hello. I played a bit with LLM Studio in H20 Aquarium while solving H2O predict LLM Competition. Models fine-tuned for Casual Classification Problems have no classification_head.pth in downloaded or pushed to HF model files. Any ideas what I have missed in the settings or why that issue happens? Thank you!

maxjeblick:
Hi, @samvelkoch, I'll have a look into this.
Could you create a new issue for better tracking? Thanks!

samvelkoch:
> Hi, @samvelkoch, I'll have a look into this. Could you create a new issue for better tracking? Thanks!

Thanks for quick reply. I've published it here since I'm not sure that it is a real issue or just my lack of experience with LLM Studio. Could you please check? I'll publish an issue now.

PR #453, opened 2023-10-17
Fix empty answers datasets

### LINKED ISSUE #451: [BUG] 0 rows after dataset import
### 🐛 Bug

I have a use-case where I wanted to pretrain llm models without q-a pairs necessarily. So, in my dataset, the prompt column is full of text but the answer column is empty strings.

When I try to import this dataset, I'm running into an error during the visualization part first and then see "train rows = 0" in the datasets table.

image

the dataset:
[oscar_tr_ds_5000_answer.csv.zip](https://github.com/h2oai/h2o-llmstudio/files/12926817/oscar_tr_ds_5000_answer.csv.zip)

To me it looks like, LLMStudio drops every row having NaN. But I thought I needed this empty column so that I could have a successful pretraining setup. (planning to select prompt -> prompt column, answer -> empty answer column and disable every prompt-token related settings)

Wonder what you think?

### LINKED ISSUE #452: [BUG] Validation datasize size may be 0
### 🐛 Bug

Using a low value for Validation Size and Data Sample size may result in a validation Dataframe with  0 rows.

### To Reproduce

- Use dataset shared in https://github.com/h2oai/h2o-llmstudio/issues/451 (5000 samples)
- Set both Validation Size and Data Sample size to 0.01

### LLM Studio version
[5d07987](https://github.com/h2oai/h2o-llmstudio/commit/5d079879e89dafda97a1620a46bc0e690ac535cf)

maxjeblick:
This PR

- Does not drop rows where the answer column is missing. Rows where the prompt column is empty will still be dropped.
- Ensures that train/val Dataframe are not empty.

Fixes  #451
Fixes #452

### DIFF
--- a/llm_studio/src/metrics/text_causal_language_modeling_metrics.py
+++ b/llm_studio/src/metrics/text_causal_language_modeling_metrics.py
@@ -28,7 +28,11 @@ def sacrebleu_score(
     for predicted_text, target_text in zip(
         results["predicted_text"], results["target_text"]
     ):
-        scores.append(metric.sentence_score(predicted_text, [target_text]).score)
+        if target_text == "":
+            score = 0.0
+        else:
+            score = metric.sentence_score(predicted_text, [target_text]).score
+        scores.append(score)
     return np.array(scores)
 
 
--- a/llm_studio/src/plots/text_causal_language_modeling_plots.py
+++ b/llm_studio/src/plots/text_causal_language_modeling_plots.py
@@ -114,7 +114,7 @@ class Plots:
         # Limit to max 15 prompt-conversation-answer rounds
         # This yields to max 5 * sum_{i=1}^{15} i = 600 rows in the DataFrame
         max_conversation_length = min(
-            max([len(conversation["prompts"]) for conversation in conversations] + [1]),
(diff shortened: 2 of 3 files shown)


llm_studio/src/metrics/text_causal_language_modeling_metrics.py
```diff
@@ -28,7 +28,11 @@ def sacrebleu_score(
     for predicted_text, target_text in zip(
         results["predicted_text"], results["target_text"]
     ):
-        scores.append(metric.sentence_score(predicted_text, [target_text]).score)
+        if target_text == "":
```
fatihozturkh2o, line 31: you sure it's gonna be still "" after reading the pandas df or NaN ?

maxjeblick: This should cast None to str: `df[input_cols + [cfg.dataset.answer_column]].fillna("").astype(str)`

fatihozturkh2o: yeah that should be fine then

fatihozturkh2o (approved):
lgtm!

PR #468, opened 2023-10-25
Classification fixes

psinger:
This PR:

- adds log loss metric for classification
- fixes errors after previous deepspeed merge

### DIFF
--- a/llm_studio/src/metrics/text_causal_classification_modeling_metrics.py
+++ b/llm_studio/src/metrics/text_causal_classification_modeling_metrics.py
@@ -4,7 +4,8 @@ from typing import Any, Dict, List, Tuple, Union
 import numpy as np
 import pandas as pd
 from numpy.typing import NDArray
-from sklearn.metrics import roc_auc_score
+from scipy.special import softmax
+from sklearn.metrics import roc_auc_score, log_loss
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


llm_studio/src/metrics/text_causal_classification_modeling_metrics.py
```diff
@@ -47,6 +62,7 @@ class Metrics:
     _metrics = {
         "AUC": (auc_score, "max", "mean"),
         "Accuracy": (accuracy_score, "max", "mean"),
+        "LogLoss": (logloss_score, "max", "mean"),
```
maxjeblick, line 65: should be `"min"` ?

maxjeblick (approved):
Thanks for the fixes, LGTM (I've left one comment).

Also, could you have a look at this line, should probably read `>0.5` ?
`preds = np.array((preds > 0.0)).astype(int).astype(str).reshape(-1)`

psinger:
> Thanks for the fixes, LGTM (I've left one comment).
> 
> Also, could you have a look at this line, should probably read `>0.5` ? `preds = np.array((preds > 0.0)).astype(int).astype(str).reshape(-1)`

it is logits, so I think `>0.0` should be correct

maxjeblick:
> it is logits, so I think >0.0 should be correct

True, it is fine then. Thanks for the clarification.

PR #471, opened 2023-10-25
Add cpu tests

maxjeblick:
This PR adds cpu-only integration tests for causal language problem type.
It uses a modified tiny llama model (https://huggingface.co/MaxJeblick/llama2-0b-unit-test). 
All tests run in ~3:30 minutes (exlucing setup).

I've removed `assert os.path.exists(os.path.join(cfg["output_directory"], "adapter_model.bin"))`, seems as if this assertion no longer works.

After this PR has been merged we can add more tests (and models) for the other problem types.


tests/integration/test_integration.py
```diff
@@ -50,50 +50,61 @@ def get_experiment_status(path: str) -> str:
...
-
     train_path = os.path.join(tmp_path, "train_full.pq")
-
-    config_path = (
-        Path.cwd() / "tests" / "src" / "integration" / f"{config_name}.yaml"
-    ).resolve()
-    with open(config_path, "r") as fp:
+    with open(
+        os.path.join(
+            os.path.dirname(os.path.realpath(__file__)), f"{config_name}.yaml"
```
pascal-pfeiffer, line 85: Is there a reason to have unit tests in `src` and the integration tests in the parent dir?

pascal-pfeiffer (commented):
Few comments, otherwise looks great! Thank you!

tests/integration/test_causal_language_modeling_oasst_cpu_cfg.yaml
```diff
@@ -0,0 +1,99 @@
...
+  mask_prompt_labels: true
+  parent_id_column: None
+  personalize: false
+  prompt_column:
+  - instruction
+  system_column: None
+  text_answer_separator: <|answer|>
+  text_prompt_start: <|prompt|>
+  text_system_start: <|system|>
+  train_dataframe: /tmp/pytest-of-max/pytest-1/test_oasst_training_cpu_BLEU_t0/train_full.pq
```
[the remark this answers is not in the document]
maxjeblick: yes, makes total sense.

tests/integration/test_integration.py
maxjeblick: The idea to move integration tests out of src was to make them more explicit and have tests in src be unit tests for the respective src folder.

pascal-pfeiffer (approved):
lgtm!

PR #477, opened 2023-10-26
More integration cpu tests

maxjeblick:
This PR adds more integration tests for CPU and GPU.

### DIFF
--- a/train.py
+++ b/train.py
@@ -569,7 +569,7 @@ def run_train_rlhf(
                     if (torch.cuda.is_available() and len(cfg.environment.gpus) > 0)
                     else torch.no_grad()
                 )
-                with context:
+                with context:  # type: ignore[attr-defined]
                     scores = reward_model.get_score(
                         batch["reward_model_prompt_text"],
                         output_dict["predicted_text"],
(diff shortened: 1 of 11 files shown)


train.py
```diff
@@ -564,7 +564,12 @@ def run_train_rlhf(
                 # tokenize prompt & output internally
                 if cfg.training.offload_reward_model:
                     reward_model.to(cfg.environment._device)
-                with autocast(enabled=cfg.environment.mixed_precision):
+                context = (
+                    autocast(enabled=cfg.environment.mixed_precision)
+                    if (torch.cuda.is_available() and len(cfg.environment.gpus) > 0)
+                    else torch.no_grad()
```
pascal-pfeiffer, line 570: Is it not possible to specify device here for autocast when using the wrapper `torch.autocast`?
It would be more intuitive to read and actually apply the setting. 

https://pytorch.org/docs/stable/amp.html
```python
# Creates model in default precision
model = Net().eval()

with torch.autocast(device_type="cpu", dtype=torch.bfloat16):
    for input in data:
        # Runs the forward pass with autocasting.
        output = model(input)
```

pascal-pfeiffer (approved):
One suggestion for code readability, otherwise lgtm. Thanks!

train.py
maxjeblick: I tried a couple of things here, but I think it would be an equally long if-else statement.

pascal-pfeiffer: not too important, thank you for checking it

PR #481, opened 2023-10-31
Refactor classification ds

maxjeblick:
Some (mild) refactoring of Causal Classification Problem Type:

- Removed some inference settings that are not needed for classification (all generation related settings). Fixed several parts code such that it will work with the changes. 
- Extracted a separate Dataset class for Causal Classification.
- Chat Tab will not be displayed when using Causal Classification.
- Fixed a (non reported) issue and removed not needed methods in the metric classes .


psinger (approved):
Thanks Max and sorry for review delay.

I couldn't find any issues, lgtm!

PR #490, opened 2023-11-09
Fix download model

### LINKED ISSUE #489: [BUG] Downloaded model may miss files
### 🐛 Bug

Some downloaded model may not work, as certain parts of the model definition (`tokenizer.model`) are missing.
See https://github.com/h2oai/h2ogpt/issues/1065#issuecomment-1803158322

### To Reproduce

1. Train a LLama2 model
2. Download the model
3. Try running it with `AutoTokenizer.from_pretrained(folder)`

 ### LLM Studio version
[f5e1448](https://github.com/h2oai/h2o-llmstudio/commit/f5e1448d5727a67ac563075cd55409e4fbe59e81)

maxjeblick:
Make Download Model functionality robust against various model backbones/tokenizers.

- Fetch tokenizer files from `tokenizer.save_pretrained`
- Monitor experiment folder and add files that where modified after the backbone/tokenizer have been saved

Fixes #489

### DIFF
--- a/llm_studio/app_utils/sections/experiment.py
+++ b/llm_studio/app_utils/sections/experiment.py
@@ -1622,7 +1622,8 @@ async def experiment_download_model(q: Q):
         model_save_time = time.time()
         model.backbone.save_pretrained(checkpoint_path)
         # See PreTrainedTokenizerBase.save_pretrained for documentation
-        # Safeguard against None return if tokenizer class is not inherited from PreTrainedTokenizerBase
+        # Safeguard against None return if tokenizer class is
+        # not inherited from PreTrainedTokenizerBase
         tokenizer_files = list(tokenizer.save_pretrained(checkpoint_path) or [])
 
         card = get_model_card(cfg, model, repo_id="<path_to_local_folder>")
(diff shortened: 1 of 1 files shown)


pascal-pfeiffer (changes requested):
`classification_head.pth` seems to be missing in the download, too, doesn't it?

maxjeblick:
Yes, you are completely right, I fixed it

pascal-pfeiffer (approved):
lgtm!


I think the model_card.md doesn't take into account the option to have multiple classes and will always assume a (1, 1) linear layer, but this is unrelated to this PR (so also wrong about input shape actually)

```
# settings can be arbitrary here as we overwrite with saved weights
head = torch.nn.Linear(1, 1, bias=False).to("cuda")
head.weight.data = head_weights

inputs = tokenizer(prompt, return_tensors="pt", add_special_tokens=False).to("cuda")

out = model(**inputs).logits

logits = head(out[:,-1])
```

maxjeblick:
> I think the model_card.md doesn't take into account the option to have multiple classes and will always assume a (1, 1) linear layer, but this is unrelated to this PR (so also wrong about input shape actually)


Yeah, I was also confused, but it is actually ok, `head.weight.data` will overwrite weights. See also the comment in the md file.

Issue #494, opened 2023-11-13
[BUG] Dataset import file refresh


psinger:
### 🐛 Bug

When importing a dataset, switching between files does not properly trigger an update on available columns.

### To Reproduce

Import a dataset with different csvs having different columns and switch between them.

Clicking continue and then back updates it correctly, so probably only the trigger somehow not working as expected.

maxjeblick:
fixed by #496

PR #496, opened 2023-11-13
fix missing ui updates

maxjeblick:
Fix for #494
It may also be possible to add a more flexible solution, would this work using `q.client["nav/active"]`?


psinger (approved):
works for me, thx

I think it theoretically could be solved with active nav, but this solutions lgtm

do we need to also trigger it for val dataframe? we wont update the columns then, or?

maxjeblick:
> do we need to also trigger it for val dataframe? we wont update the columns then, or?

Good point, removed it

Issue #495, opened 2023-11-13
[FEATURE] Causal Regression Modeling


psinger:
### 🚀 Feature

Similar to classification, we can support causal regression as a new problem type.

PR #507, opened 2023-11-24
improve problem type handling

maxjeblick:
Improve handling for non generating problem types (see also #495).


psinger (approved):
Great - thx!

PR #509, opened 2023-11-29
Checkpoint logging fix

psinger:
Fixing tiny inconsistency across ranks for logging.

Also icnreasing slider for inference length.


maxjeblick (approved):
LGTM thanks!

PR #513, opened 2023-12-01
fix deepspeed w/ lora generate speed

### LINKED ISSUE #461: [CODE IMPROVEMENT] Improve deepspeed w/ lora generate speed

haqishen:
close #461

I found that it's actually no need to merge & unmerge the adapter weights into the LoRA layers before & after generation, because PEFT has two parameters that control whether to use the adapter weights in the forward flow. One is `merged` (as we discussed earlier, if we merge weights, it will be set to True), and the other is `disable_adapters` (as shown in https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/layer.py#L354-L370 ). For our purposes, these two are completely equivalent, while we don't need to deal with `disable_adapters` either, cause it's `False` during the whole process.

Exp results shown below have also confirmed this.

# Exp Result

| Backbone                    | Method                                  | Train Epochs | Eval GPT Metric  |
|-----------------------------|-----------------------------------------|-------|------------------|
| h2oai/h2ogpt-4096-llama2-7b | ddp w/ lora                                         | 0     | 0.0              |
| h2oai/h2ogpt-4096-llama2-7b | ddp w/ lora                        | 1     | around 0.5       |
| h2oai/h2ogpt-4096-llama2-7b | deepspeed w/ lora (latest commit)         | 1     | around 0.5       |
| h2oai/h2ogpt-4096-llama2-7b | deepspeed w/ lora (latest commit, push ↑ to HF and pull) | 0   | around 0.5       |

# Deepspeed Generation Speed

10x~20x faster than before.


llm_studio/src/models/text_causal_language_modeling_model.py
```diff
@@ -55,6 +56,24 @@ def init_deepspeed(self):
...
+                    pass
+
+    def unmerge_adapter(self):
+        for module in self.backbone.modules():
+            if isinstance(module, BaseTunerLayer):
+                # when using deepspeed, not modules layers are in all ranks
+                try:
+                    module.unmerge()
+                except:
+                    pass
```
haqishen, line 75: https://github.com/huggingface/peft/blob/6a57472665b2b712a84e2bedd98945038283f7cc/src/peft/tuners/tuners_utils.py#L273-L287

modified from here.

pascal-pfeiffer (changes requested):
Thanks a lot for the PR @haqishen 
Great improvement on the generate speed. On my tests, it was about 50% of the original time. (70B llama2 model)
Can you please fix the typing issues and comment on the changes that you made to the default settings? Otherwise lgtm, thanks again.

pascal-pfeiffer:
> 10x~20x faster than before.

Where did you see that huge speedup? I can confirm about twice the speed. So, 50% time for inference vs main branch.

haqishen:
@pascal-pfeiffer Thanks for the comments!

After fixing this issue, these three parameters will start to affect the generation speed. The larger they are set, the faster the speed will be. When I tested llama2 7B on 3xRTX6000, a setting of 1e9 allowed the generation speed to reach its peak, which is a tenfold increase compared to the original.

I am currently testing the maximum parameter settings for llama2 70B with 4xA6000. We then set the default parameters to this value. What do you think?

pascal-pfeiffer:
> @pascal-pfeiffer Thanks for the comments!
> 
> After fixing this issue, these three parameters will start to affect the generation speed. The larger they are set, the faster the speed will be. When I tested llama2 7B on 3xRTX6000, a setting of 1e9 allowed the generation speed to reach its peak, which is a tenfold increase compared to the main branch.
> 
> I am currently testing the maximum parameter settings for llama2 70B with 4xA6000. We then set the default parameters to this value. What do you think?

The RTX A6000 doesn't have NVLink, so any sharding will probably be bound by the inter GPU connection. I am not exactly sure how they translate to the shards. With 1e9, maybe it is just loading the full model to each card?

I'd really leave it at the original default (which worked for V100 cards) and expand the tooltip that this should be tuned as large as possible for the given memory. Unless there is a reliable automatic way to find optimum values.

psinger:
One thing I noticed to not work properly is a check for enough disk space before starting an experiment.

Getting `OSError: [Errno 28] No space left on device` while deepspeed saves weights.

haqishen:
> One thing I noticed to not work properly is a check for enough disk space before starting an experiment.
> 
> Getting `OSError: [Errno 28] No space left on device` while deepspeed saves weights.

In the previous PR, there was a commit (https://github.com/h2oai/h2o-llmstudio/pull/288/commits/62fc9c5052076c455b4fa3affe027c453bd0d8c5)  that saved the model directly in PyTorch format using Deepspeed. Later, due to suspicions that this code caused the model accuracy reduction issue mentioned by Judan, it was reverted. 

I will retest this code to try to resolve the issue you are encountering.

llm_studio/src/utils/modeling_utils.py
```diff
@@ -58,21 +58,24 @@ def check_disk_space(model: torch.nn.Module, path: str, use_deepspeed: bool = Fa
...
-            model_size_in_bytes += param.numel() * 4
+            model_size_in_bytes += n_params * 4
         else:
             # If the data type is not supported, calculate it as float32.
-            model_size_in_bytes += param.numel() * 4
+            model_size_in_bytes += n_params * 4
             logger.warning(f"Unsupported data type: {param.data.dtype}")
 
-    if use_deepspeed:
-        model_size_in_bytes *= 2  # need double space for converting deepspeed engine.
```
haqishen, line 73: by saving pytorch checkpoint directly for deepspeed, we don't need to `x2` for space check now.

llm_studio/src/utils/modeling_utils.py
```diff
@@ -58,21 +58,24 @@ def check_disk_space(model: torch.nn.Module, path: str, use_deepspeed: bool = Fa
 
     model_size_in_bytes = 0
     for param in model.parameters():
+        n_params = param.ds_numel if hasattr(param, "ds_numel") else param.numel()
```
haqishen, line 61: fix bug for check disk space when using deepspeed

llm_studio/src/utils/modeling_utils.py
```diff
@@ -95,17 +98,30 @@ def save_checkpoint(model: torch.nn.Module, path: str, cfg: Any):
...
     if cfg.environment.use_deepspeed:
         if path is not None:
-            # gather model params from all ranks
-            model.save_checkpoint(os.path.join(path, "ds_checkpoint"))  # type: ignore[operator] # noqa: E501
-            if cfg.environment._local_rank == 0:
-                # load to cpu
-                state_dict = get_fp32_state_dict_from_zero_checkpoint(
-                    os.path.join(path, "ds_checkpoint")
+            # gather model params from all ranks when using Deepspeed
+            status = model.save_16bit_model(path, "checkpoint.pth")
```
haqishen, line 102: saving pytorch checkpoint directly for deepspeed

haqishen:
> The RTX A6000 doesn't have NVLink, so any sharding will probably be bound by the inter GPU connection. I am not exactly sure how they translate to the shards. With 1e9, maybe it is just loading the full model to each card?

Yeah, I think there's a high probability that this is the case. I have reverted the default values and made slight change to the corresponding doc. Please check: [f378fd8](https://github.com/h2oai/h2o-llmstudio/pull/513/commits/f378fd8dcb649668ea827da1742cbfda06842fdb)

haqishen:
> One thing I noticed to not work properly is a check for enough disk space before starting an experiment.
> 
> Getting `OSError: [Errno 28] No space left on device` while deepspeed saves weights.

Hi @psinger I've fixed the check disk space and save checkpoint issue you met and checked the metric of train-push-pull experiments, please try again! [c60ab87](https://github.com/h2oai/h2o-llmstudio/pull/513/commits/c60ab87b74cddd2ebe12e5cddafbeaab717a28ae)

pascal-pfeiffer (approved):
thanks for the fixes @haqishen 
lgtm !

PR #525, opened 2023-12-04
Munish/UI tests

itsmunishbhardwaj:


tests/ui/llm_studio.feature
```diff
@@ -0,0 +1,19 @@
+Feature: LLM Studio
+
+    Background: LLM Studio user
+        Given LLM Studio home page is opened
+        When I login to LLM Studio
+        Then I see the home page
+
+    Scenario: Import dataset using filesystem
+        When I upload dataset with path /home/llmstudio/mount/data/user/oasst/train_full.pq and name test_dataset
```
mnezh, line 9: looks like deployment-specific file path. could it be something like:
```gherkin
When I upload dataset oasst/train_full.pq as test_dataset
```
And default location could different for local and remove test execution?

tests/ui/test.py
```diff
@@ -0,0 +1,48 @@
...
+
+@given("LLM Studio home page is opened")
+def open_llm_studio(page: Page, app_address: str):
+    page.goto(app_address)
+    handle_terms_and_conditions_page(page)
+
+
+@when("I login to LLM Studio", target_fixture="llm_studio")
+def login_to_llm_studio(logger: logging.Logger, page: Page, app_address: str):
+    login(page, "okta", "H2O-Tester", "ManagedCloud5")
```
mnezh, line 18: don't use hardcoded login type or credentials

tests/ui/utils.py
```diff
@@ -0,0 +1,140 @@
...
+    terms_and_conditions = page.get_by_role("heading", name="EULA")
+
+    if terms_and_conditions.is_visible():
+        # If the heading is present, click the "I agree" button
+        page.get_by_role("button", name="I agree").click()
+    else:
+        return 1
+
+
+class LLMStudioPage(BasePage):
```
mnezh, line 34: please separate utils and a page object class into different files

tests/ui/utils.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def open_home_page(self):
+        self.page.get_by_role("button", name="Home").click()
+
+    def open_app_settings(self):
+        self.page.get_by_role("button", name="Settings").click()
+
+    def import_dataset_from_filesystem(self, path: str, filename: str):
+        self.page.get_by_role("button", name="Import dataset").click()
+        self.page.locator('[data-test="dataset\\/import\\/source"]').get_by_text(
```
mnezh, line 60: please use https://playwright.dev/python/docs/locators#locate-by-test-id and everywhere else

Makefile
```diff
@@ -25,6 +26,11 @@ setup: pipenv
 setup-dev: pipenv
 	$(PIPENV) install --verbose --dev --python $(PYTHON_VERSION)
 
+setup-ui-test:
```
mnezh, line 29: please use `[dev-packages]` for that purpose https://pipenv.pypa.io/en/latest/pipfile.html
the team is using pipenv, let's be consistent and extend:
https://github.com/h2oai/h2o-llmstudio/blob/main/Pipfile#L56

mnezh (changes requested):

Makefile
mnezh: @psinger do you have an idea why @itsmunishbhardwaj might have the problem above?

psinger: not really, but we mainly support Linux and not Mac

maxjeblick: FYI, mac is not supported

PR #531, opened 2023-12-07
Add neftune

### LINKED ISSUE #492: [FEATURE] Support NEFTune noisy embedding vectors technique
### 🚀 Feature

In this [paper](https://arxiv.org/pdf/2310.05914.pdf), the authors introduced a concept they called NEFTune, which basically shows that LLM downstream performance can be improved if noise is added to the embedding vectors during instruct fine-tuning. Could this capability be supported in LLM Studio? The capability has already been added to Huggingface's TRL (Transformer Reinforcement Library).

### Motivation

The promising results of the paper suggest it could be useful to have the ability to have this capability in LLM Studio to maximize downstream performance of LLMs trained in the platform.

Here is a link to their [repo](https://github.com/neelsjain/NEFTune).

maxjeblick:
This PR adds neftune as proposed by _NEFTune: Noisy Embeddings Improve Instruction Finetuning https://arxiv.org/abs/2310.05914._

I've tested the functionality manually with ddp and deepspeed enabled/disabled.

@tmostak reported some improvements while testing this feature on this branch (see #492)

Fixes #492

### DIFF
--- a/documentation/docs/faqs.md
+++ b/documentation/docs/faqs.md
@@ -2,7 +2,7 @@ import Icon from "@material-ui/core/Icon";
 
 # FAQs
 
-The sections below provide answers to frequently asked questions. If you have additional questions, please send them to <cloud-feedback@h2o.ai>.
+The sections below provide answers to frequently asked questions. If you have additional questions, please send them to [cloud-feedback@h2o.ai](mailto:cloud-feedback@h2o.ai).
 
 ---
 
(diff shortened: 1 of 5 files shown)


pascal-pfeiffer (approved):
Thank you @maxjeblick , very clean code
lgtm

we should check for potential needed changes in https://github.com/h2oai/h2o-llmstudio/pull/513 , though it currently works flawless with deepspeed, too.

PR #549, opened 2023-12-20
UI Automation tests for LLM Studio

itsmunishbhardwaj:
### Framework
We have used Python BDD Framework with playwright.

### Tree
```
llm_studio/
|-- tests/
|   |-- ui/
|       |-- __init__.py
|       |-- conftest.py
|       |-- utils.py
|       |-- llm_studio_page.py
|       |-- llm_studio.feature
|       |-- test.py
```

### How to run the tests
- run the latest version for llm studio app on cloud platforms like https://internal-genai.dedicated.h2o.ai/
- in the console export following environment variables:
  ```bash
  export APP_URL=
  export OKTA_USER=
  export OKTA_PASSWORD=
  ```
- setup environment
  ```bash
  make setup-ui-test
  ```
- run the tests
  ```bash
  make test-ui
  ```

> Please note that this testing framework requires an instance to running prior. To integrate this test suite to CI, I will need your help.

### Important Points for Discussion:
- For now I have not added separate logging to the codebase. Should I add logging? @mnezh 
- Theres a method in playwright called `get_by_test_id` which helps testing by test ids ie:
   - for http elements like:
      ``` http
      <button data-testid="directions">Itinéraire</button>
      ```
    - we can simply do:
      ```
      page.get_by_test_id("directions").click()
      ```
    - However we dont have `data-testid=` in UI codebase. We instead have `data-test`.
    - The good news is there is a method to tackle this https://playwright.dev/python/docs/locators#set-a-custom-test-id-attribute But due to lack of documentation I wasnt able to use it
    - So I created my own method by the same name.
    - Can you please help me tackle this? @mnezh 

- Should I also add piplock to the repository? @maxjeblick


maxjeblick:
Thanks a lot for the RR @itsmunishbhardwaj ! I will have a look at it the next days.

> Should I also add piplock to the repository? @maxjeblick
Yes, this is needed; I created a commit in this branch to add `Pipfile.lock`.

Makefile
```diff
@@ -31,6 +31,10 @@ setup-dev: pipenv
 setup-no-flash: pipenv
 	$(PIPENV) install --verbose --python $(PYTHON_VERSION)
 
+setup-ui-test: pipenv
+	$(PIPENV) install --verbose --categories=dev-packages  
+	$(PIPENV) run playwright install
```
maxjeblick, line 36: Could you also add playwright install to `setup-dev` so that `setup-dev` installs all dependencies needed for testing?

tests/ui/llm_studio_page.py
```diff
@@ -0,0 +1,242 @@
...
+    DATASET_DELETE_DIALOG_SELECTOR = "dataset/delete/dialog"
+    DATASET_DELETE_SELECTOR = "dataset/delete"
+    EXPERIMENT_START_RUN_SELECTOR = "experiment/start/run"
+    EXPERIMENT_NAME_SELECTOR = "experiment/start/cfg/experiment_name"
+    EXPERIMENT_LIST_DELETE_SELECTOR = "experiment/list/delete"
+    EXPERIMENT_DELETE_DIALOG_SELECTOR = "experiment/delete/dialog"
+    EXPERIMENT_DELETE_SELECTOR = "experiment/delete"
+    EXPERIMENT_STATUS_SELECTOR = "[data-automation-key='status']"
+    EXPERIMENT_INDEX_SELECTOR = "[data-automation-key='name']"
+    FILESYSTEM_PATH = "/home/llmstudio/mount/data/user/"
```
maxjeblick, line 19: This filepath won't work for a locally running instance.

tests/ui/llm_studio_page.py
```diff
@@ -0,0 +1,242 @@
...
+                self.get_by_test_id(self.EXPERIMENT_REFRESH_SELECTOR).click()
+                status = self.page.locator(
+                    f"{self.EXPERIMENT_STATUS_SELECTOR} >> nth={i}"
+                ).inner_text()
+            elif status == "finished":
+                break
+
+    def find_experiment_index(self, experiment_name):
+        i = 0
+        while True:
```
maxjeblick, line 211: Looks risky

tests/ui/llm_studio.feature
```diff
@@ -0,0 +1,30 @@
+Feature: LLM Studio
+
+    Background: LLM Studio user
```
maxjeblick, line 3: That test would need to be excluded from non-cloud ui tests.

Makefile
```diff
@@ -84,6 +88,18 @@ test: reports
...
+.PHONY: test-ui
+test-ui: reports
+	$(PIPENV) run pytest -v --junitxml=reports/junit_ui.xml \
+	--html=./reports/pytest_ui.html \
+	-o log_cli=true -o log_level=INFO -o log_file=reports/tests_ui.log \
+	tests/ui/test.py 2>&1 | tee reports/tests_ui.log
+
+.PHONY: test-ui-local
+test-ui-local: 
+	$(PIPENV) run pytest -vvs --headed \
```
mnezh, line 100: same about logging

tests/ui/conftest.py
```diff
@@ -0,0 +1,27 @@
...
+logging.Logger.trace = partialmethod(logging.Logger.log, logging.TRACE)
+logging.trace = partial(logging.log, logging.TRACE)
+
+
+@pytest.fixture(scope="session")
+def logger() -> logging.Logger:
+    return logging.getLogger("ui-tests")
+
+
+@pytest.fixture(scope="session")
```
mnezh, line 18: already implemented https://playwright.dev/python/docs/test-runners#configure-base-url

tests/ui/llm_studio.feature
```diff
@@ -0,0 +1,30 @@
...
+        Given LLM Studio home page is opened
+        When I login to LLM Studio
+        Then I see the home page
+
+    Scenario: Import dataset using filesystem
+        When I upload dataset with path oasst/train_full.pq and name test-dataset
+        Then I should see the dataset test-dataset
+
+    Scenario: Delete dataset
+        When I delete dataset test-dataset
```
mnezh, line 13: scenario starts assuming there's a dataset already, i.e. depends on previous scenario.
in general, that's not a good idea. if you need a flow, join to scenarios.
if you insist on having two scenarios, make them independent on the outcome of each other.

tests/ui/llm_studio.feature
```diff
@@ -0,0 +1,30 @@
...
+        And I update LLM Backbone to MaxJeblick/llama2-0b-unit-test
+        And I tweak data sampling to 0.1
+        And I tweak max length prompt to 128
+        And I tweak max length answer to 128
+        And I tweak max length to 32
+        And I run the experiment
+        Then I should see the test-experiment should finish successfully 
+        
+    
+    Scenario: Delete experiment
```
mnezh, line 27: same as above

tests/ui/llm_studio_page.py
mnezh: 1. this should be passed from the outside of the tests and be part of test configuration
2. the file path doesn't belong to the Page Object, which is the abstraction level for the UI actions, but rather to the test itself

tests/ui/llm_studio_page.py
```diff
@@ -0,0 +1,242 @@
...
+            if value_now == target_value:
+                is_completed = True
+            else:
+                # Move the slider a little bit (adjust the step as needed)
+                step = 0.1  # Adjust this value based on your requirements
+                x1 = x2
+                i += step
+
+    def run_experiment(self):
+        self.page.wait_for_timeout(3000)
```
mnezh, line 160: 1. let's use a constant with a meaningful name, if we need need a shorter timeout than the default one
2. why do we need a shorter timeout? for a quick test failure?

tests/ui/llm_studio_page.py
```diff
@@ -0,0 +1,242 @@
...
+        locator.click()
+
+    def view_experiment(self, experiment_name: str):
+        self.view_experiment_page()
+        i = self.find_experiment_index(experiment_name)
+        status = self.page.locator(
+            f"{self.EXPERIMENT_STATUS_SELECTOR} >> nth={i}"
+        ).inner_text()
+
+        while True:
```
mnezh, line 199: we can wait for locator with filters instead of re-implementing the poll loop
https://playwright.dev/python/docs/locators#filtering-locators

mnezh (changes requested):

ddeloy:
LGTM as well

Makefile
```diff
@@ -84,6 +88,18 @@ test: reports
     -o log_cli=true -o log_level=INFO -o log_file=reports/tests.log \
     tests/* 2>&1 | tee reports/tests.log'
 
+.PHONY: test-ui
+test-ui: reports
+	$(PIPENV) run pytest -v --junitxml=reports/junit_ui.xml \
```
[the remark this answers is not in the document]
itsmunishbhardwaj: done!

tests/ui/conftest.py
itsmunishbhardwaj: so should I remove this block and add --base-url flag to the make command??

tests/ui/llm_studio.feature
itsmunishbhardwaj: All ui tests are cloud only ui tests @maxjeblick. This PR only addresses UI testing on Cloud

tests/ui/llm_studio.feature
itsmunishbhardwaj: Oh intresting, lets combine them in that case @mnezh

tests/ui/llm_studio_page.py
itsmunishbhardwaj: I was trying to implement sleep here @mnezh. This UI was crashing randomly at this point while tests locally

itsmunishbhardwaj: Let me find a better way to do this

tests/ui/conftest.py
mnezh: yes, unless we have a serious reason for reimplementing existing functionality

=2023.11.15
```diff
@@ -0,0 +1,7 @@
+Requirement already satisfied: pipenv in /opt/hostedtoolcache/Python/3.10.11/x64/lib/python3.10/site-packages (2023.11.15)
```
mnezh, line 1: this file is probably a temporary leftover and should be removed?

mnezh (changes requested):

tests/ui/llm_studio_page.py
mnezh: @itsmunishbhardwaj if the UI is crashing, probably it needs some bugfix? Could you please investigate the bug and raise a ticket for @maxjeblick  to review?

itsmunishbhardwaj: Its a random event @mnezh Im unable to capture it. Theres a "sorry for inconvenience" message that appears during a crash

=2023.11.15
itsmunishbhardwaj: Im unable to delete this file @mnezh please help 🙏🏻

mnezh:
```
git rm ‘=2023.11.15’
git add -A
git commit -m ‘remove file =2023.11.15’
```

itsmunishbhardwaj:
Duplicate https://github.com/h2oai/h2o-llmstudio/pull/561

PR #552, opened 2023-12-21
MT-BENCH Metric

psinger:
This PR adds a new functionality to calculate a (proxy) for mt-bench metric based on https://github.com/lm-sys/FastChat/tree/main/fastchat/llm_judge directly in LLM Studio.

Workflow:
- Choose mt-bench GPT template
- Select `gpt-4-0613` endpoint
- use 1024 max output token length
- Internally, the val dataset will then be replaced by 80 questions from mt-bench
- We run typical generation on those
- A re-implementation of mt-bench templates will be run, separately for 30 questions with reference and 50 without reference answer - slight adaptations to the templates are made to match our output format
- Score will be shown, as well as scores by category in the logs

Some tests show good match between mt-bench and our implementation. For example for `mistralai/Mistral-7B-Instruct-v0.2` I get 7.49 on mt-bench, and 7.44 in LLM Studio:

```
2023-12-21 09:33:15,910 - INFO: MT-Bench scores by category:
  category  score
    coding   5.80
extraction   5.10
humanities   9.75
      math   3.80
 reasoning   6.00
  roleplay   9.30
      stem   9.90
   writing   9.90
2023-12-21 09:33:15,911 - INFO: Validation GPT: 7.44375
```

Some differences vs. the original implementation:

- mt-bench uses different temperatures per category: https://github.com/lm-sys/FastChat/blob/main/fastchat/llm_judge/common.py#L36
- tiny template differences (also for sys prompt to openai)
- mt-bench calculates both first and second turn, we only do first turn

Some potential considerations:

- Do not automatically replace the val dataframe, but user needs to prep, but I like the simplicity of automatically providing it, so I would keep it like that for now. 
- Force the gpt endpoint for mt-bench, but might be also limiting and higher costs


pascal-pfeiffer (changes requested):
Great addition, thank you! 

Let's add some more information about how this metric can be used in the tooltip. Basically what you described in the comment to this PR. Then we can merge and keep two tasks as follow up:

- update docs
- implement the second turn eval

pascal-pfeiffer:
We should also make clear in the UI and logs, that the result will differ from the original implementation of MT-bench to avoid confusion.

psinger:
> We should also make clear in the UI and logs, that the result will differ from the original implementation of MT-bench to avoid confusion.

Sure, but the whole metric has a random range by default, so not sure too many disclaimers are really needed.

pascal-pfeiffer (approved):
Thank you, lgtm!

PR #556, opened 2023-12-28
Enhancing Length Consistency in LLM Outputs with Token Length Penalty Loss Functions

Nischaydnk:
Adding support for custom loss functions aimed at improving the length consistency in responses generated by fientuned LLMs. Idea is to make the output lengths of LLMs more reflective of the token lengths observed in the training data. I did several experiments using the loss functions, and noticed very low deviation in performance of models. 

The loss functions implemented are:

LengthBasedTACE (Token Averaged Cross Entropy)
LengthBasedSACE (Sample Averaged Cross Entropy)

Sharing some of the experiments I did using these losses to make a comparison with original Cross Entropy Loss:

#### Evaluation Results:

There could be some randomness involved in eval metric, but I found consistent decrease in LLMs inference time,specially the ones which scores bad & prone to generate bad responses.

| Model            | Loss Function                         | Time Taken (min) | Eval Metric |
|------------------|---------------------------------------|------------------|-------------------|
| llama13B-Chat    | Token Avg CE Loss                     | 40.45               | 0.810            |
| llama13B-Chat    | TokenLengthPenalty Token Avg          | 38.62               | 0.802            |
| llama7B-Chat    | Token Avg CE Loss                     | 12.50               | 0.7684            |
| llama7B-Chat    | TokenLengthPenalty Token Avg          | 12.12               | 0.7484            |
| Yi-6B-Chat    | Token Avg CE Loss                     | 18.50               | 0.792             |
| Yi-6B-Chat    | TokenLengthPenalty Token Avg          | 15.44               | 0.785             |
| llama13B-Chat    | Token Avg CE Loss                     | 78.20               | 0.728             |
| llama13B-Chat    | TokenLengthPenalty Token Avg          | 76.60               | 0.744             |
|   Yi-6B-Chat  | Token Avg CE Loss                     | 24.44               | 0.712             |
| Yi-6B-Chat    | TokenLengthPenalty Token Avg          | 24.20               | 0.704             |

These functions uses a length penalty coefficient, in my experiments I found 0.1 coefficient to be most stable one, therefore I kept it as default. This should help close #537

PR #559, opened 2024-01-02
Length Consistency in LLM Outputs with Token Length based Penalty Loss Functions

### LINKED ISSUE #537: [FEATURE] Pack sequences in batch
### 🚀 Feature

When fine-tuning a common thing to happen is that samples have very different length, particularly when we chain longer conversations together. It is a regular practice to pack sequences together in a batch, to fill up the vailable size.

The current solution is inefficient particularly in multi-gpu setups.

I have not fully explored what the best strategy is here. Can also have a look how other libraries are doing it, such as:
https://github.com/OpenAccess-AI-Collective/axolotl/blob/450e04d3c460828be66937426a91cfd161973a87/src/axolotl/utils/samplers/multipack.py#L105

Additionally, we should explore if scaling the loss based on actual tokens per sample trained on is something to look into.

Nischaydnk:
Reopened the same PR #556 with correct local branch as suggested by @psinger. 

Adding support for custom loss functions aimed at improving the length consistency in responses generated by fientuned LLMs. Idea is to make the output lengths of LLMs more reflective of the token lengths observed in the training data. I did several experiments using the loss functions, and noticed very low deviation in performance of models. 

The loss functions implemented are:

LengthBasedTACE (Token Averaged Cross Entropy)
LengthBasedSACE (Sample Averaged Cross Entropy)

Sharing some of the experiments I did using these losses to make a comparison with original Cross Entropy Loss:

#### Evaluation Results:

There could be some randomness involved in eval metric, but I found consistent decrease in LLMs inference time,specially the ones which scores bad & prone to generate bad responses.

| Model            | Loss Function                         | Time Taken (min) | Eval Metric |
|------------------|---------------------------------------|------------------|-------------------|
| llama13B-Chat    | Token Avg CE Loss                     | 40.45               | 0.810            |
| llama13B-Chat    | TokenLengthPenalty Token Avg          | 38.62               | 0.802            |
| llama7B-Chat    | Token Avg CE Loss                     | 12.50               | 0.7684            |
| llama7B-Chat    | TokenLengthPenalty Token Avg          | 12.12               | 0.7484            |
| Yi-6B-Chat    | Token Avg CE Loss                     | 18.50               | 0.792             |
| Yi-6B-Chat    | TokenLengthPenalty Token Avg          | 15.44               | 0.785             |
| llama13B-Chat    | Token Avg CE Loss                     | 78.20               | 0.728             |
| llama13B-Chat    | TokenLengthPenalty Token Avg          | 76.60               | 0.744             |
|   Yi-6B-Chat  | Token Avg CE Loss                     | 24.44               | 0.712             |
| Yi-6B-Chat    | TokenLengthPenalty Token Avg          | 24.20               | 0.704             |

These functions uses a length penalty coefficient, in my experiments I found 0.1 coefficient to be most stable one, therefore I kept it as default. This should help close #537


llm_studio/src/losses/text_causal_language_modeling_losses.py
```diff
@@ -25,6 +24,33 @@ def forward(self, logits, labels):
...
+        shift_logits = logits[..., :-1, :].contiguous()
+        shift_labels = labels[..., 1:].contiguous()
+
+        shift_logits = shift_logits.view(-1, shift_logits.size(-1))
+        shift_labels = shift_labels.view(-1)
+
+        loss = self.loss_fn(shift_logits, shift_labels)
+
+        true_lengths = torch.sum(labels != 0, dim=-1).float()
+        pred_lengths = torch.sum(torch.argmax(logits, dim=-1) != 0, dim=-1).float()
```
psinger, line 44: why exactly are we checking for `0` here?

Nischaydnk: The idea was to ignore the PAD token Ids which we generally apply on answers text (max_length_answer). In models like Llama2, etc. Pad token id is 0, so I took it as default. But yeah, it could be possible that token ID may differ for different tokenizers.

psinger: Yeah this will not be the case all the time. Actually, most of the time we set it to eos token. This needs to be dynamically set. Actually, it might be better to use the `attention_mask` for it. Similar logic like in batch padding where we are also calculating the length: https://github.com/h2oai/h2o-llmstudio/blob/8d1c13684bc5e173b635e4d7867fa84d31c5998e/llm_studio/src/utils/data_utils.py#L640

psinger (changes requested):
See comment, need to handle the length estimation better

llm_studio/src/losses/text_causal_language_modeling_losses.py
Nischaydnk: Yes right, Thanks for pointing out. I will update the loss function based on the tokenizer.

psinger: Most should be in the `cfg` already stored, can look at the `get_tokenizer` function.

psinger:
@Nischaydnk will you find time to continue working on this?

psinger:
closing this for now - please re-open in future

PR #561, opened 2024-01-12
UI Automation Tests for LLM Studio

itsmunishbhardwaj:
# How to run the tests

- To run tests locally 
  ```bash
   export LOCAL_LOGIN=True
   export PYTEST_BASE_URL=localhost:10101
   make llmstudio
   make test-ui
  ```
- To run tests on platform with okta login like:  https://internal-genai.dedicated.h2o.ai/
  ```bash
   export OKTA_USER=
   export OKTA_PASSWORD=
   export PYTEST_BASE_URL=
   make test-ui
  ```
- To run tests on platform with keycloak login like: https://cloud-qa.h2o.ai/
  ```bash
   export KEYCLOAK_USER=
   export KEYCLOAK_PASSWORD=
   export PYTEST_BASE_URL=
   make test-ui
  ```

Test results are stored at `reports/junit_ui.xml`


tests/ui/utils.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def handle_terms_and_conditions_page(page: Page):
+    terms_and_conditions = page.get_by_role("heading", name="EULA")
+
+    if terms_and_conditions.is_visible():
+        # If the heading is present, click the "I agree" button
+        page.get_by_role("button", name="I agree").click()
+    else:
+        return 1
```
mnezh, line 29: 1. this method has None return type
2. the return value is not used anywhere in the code
thus entire section could be removed:
```
else:
        return 1
```

mnezh (changes requested):

Makefile
```diff
@@ -26,11 +32,17 @@ setup: pipenv
 setup-dev: pipenv
 	$(PIPENV) install --verbose --dev --python $(PYTHON_VERSION)
 	- $(PIPENV_PIP) install flash-attn==2.3.6 --no-build-isolation --upgrade
+	- $(PIPENV_PIP) install playwright==1.40.0
```
maxjeblick, line 35: playwright should be installed already, see pipfile?

maxjeblick (changes requested):
I still have issues running the tests locally (`make test-ui-local`), in particular it fails on `page.goto(base_url)` with `base_url` being an empty string.

tests/ui/test.py
```diff
@@ -0,0 +1,102 @@
+import logging
+import os
+
+from playwright.sync_api import Page
+from pytest_bdd import given, parsers, scenarios, then, when
+from ui.llm_studio_page import LLMStudioPage
```
maxjeblick, line 6: Could you change this line (and also other imports) a relative import, i.e. `from .llm_studio_page import`?

tests/ui/llm_studio_page.py
```diff
@@ -0,0 +1,241 @@
+from hac_playwright.pages.base import BasePage
+from playwright.sync_api import expect
+
+CLOUD_FILESYSTEM_PATH = "/home/llmstudio/mount/data/user/"
```
maxjeblick, line 4: As discussed, this won't work for local tests.
Probably an env variable in the makefile together with an if-else condition will do the job (unless there's a more direct way of detecting running on the cloud):
```
test-ui-local: 
	$LOCAL_TEST=true \
        (PW_DEBUG) $(PIPENV) run pytest \ ...

```
Within the python code:
`if "LOCAL_TEST" in os.environ`

itsmunishbhardwaj: I can create an env variable to store the local directory to run the tests locally. WDYT? eg:
`export FILESYSTEM_PATH=<path to local filesystem>`
you can manually export this env var to add path which shall trump `$CLOUD_FILESYSTEM_PATH`

maxjeblick: There may be other places where one may need to distinguish between local vs. cloud (e.g. disable login for local), having a flag may be more flexible.

itsmunishbhardwaj: makes sense yess

tests/ui/llm_studio.feature
```diff
@@ -0,0 +1,25 @@
...
+        And I update LLM Backbone to MaxJeblick/llama2-0b-unit-test
+        And I tweak data sampling to 0.5
+        And I tweak max length prompt to 128
+        And I tweak max length answer to 128
+        And I tweak max length to 32
+        And I run the experiment
+        Then I should see the test-experiment should finish successfully 
+        When I delete experiment test-experiment
+        Then I should not see the experiment test-experiment
+
```
mnezh, line 25: Line 25 contains spaces "        ", but it should be just new line character at the end of 24

tests/ui/llm_studio_page.py
mnezh: In general, Page Objects should not contain any test logic and test data, like specific paths, names etc.
It's just an abstraction layer to express your tests steps.
So all paths, names belong to test cases/fixtures only.

tests/ui/llm_studio_page.py
```diff
@@ -0,0 +1,245 @@
...
+            else:
+                # Move the slider a little bit (adjust the step as needed)
+                step = 0.1  # Adjust this value based on your requirements
+                x1 = x2
+                i += step
+
+    def run_experiment(self):
+        locator = self.get_by_test_id(self.EXPERIMENT_RUN_SELECTOR)
+        locator.click()
+        self.page.wait_for_timeout(3000)
```
mnezh, line 167: let's wait for specific selector (i.e. DOM change) instead of timeout.

tests/ui/llm_studio_page.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def max_length(self, value):
+        self.slider(self.MAX_LENGTH, value)
+
+    def max_length_inference(self, value):
+        self.slider(self.MAX_LENGTH_INFERENCE, value)
+
+    def view_experiment_page(self):
+        locator = self.page.get_by_role("button", name="View experiments")
+        locator.click()
```
mnezh, line 194: 1. self.page.get_by_role("button", name="View experiments").click()
2. shouldn't we wait for the experiments page to be displayed?

tests/ui/llm_studio_page.py
```diff
@@ -0,0 +1,245 @@
...
+                x1 = x2
+                i += step
+
+    def run_experiment(self):
+        locator = self.get_by_test_id(self.EXPERIMENT_RUN_SELECTOR)
+        locator.click()
+        self.page.wait_for_timeout(3000)
+
+    def experiment_name(self, name: str):
+        locator = self.get_by_test_id(self.EXPERIMENT_NAME_SELECTOR)
```
mnezh, line 170: self.get_by_test_id(self.EXPERIMENT_NAME_SELECTOR).fill(name)

tests/ui/llm_studio_page.py
```diff
@@ -0,0 +1,245 @@
...
+        locator.fill(name)
+
+    def llm_backbone(self, value: str):
+        locator = self.page.get_by_role("combobox", name="LLM Backbone")
```
mnezh, line 174: self.page.get_by_role("combobox", name="LLM Backbone").fill(value)

tests/ui/test.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+@given("LLM Studio home page is opened")
+def open_llm_studio(page: Page, base_url):
+    page.goto(base_url)
+
+
+@when("I login to LLM Studio", target_fixture="llm_studio")
+def login_to_llm_studio(logger: logging.Logger, page: Page, base_url: str):
+    okta_user = os.environ.get("OKTA_USER")
```
mnezh, line 19: what if we deploy LLM Studio to GCP env with keycloak instead of Okta?

mnezh (changes requested):

tests/ui/llm_studio_page.py
itsmunishbhardwaj: @mnezh you asked me to remove filesystem path from the testfile

tests/ui/llm_studio_page.py
itsmunishbhardwaj: We dont have to

itsmunishbhardwaj: Is it failing for you?

tests/ui/test.py
itsmunishbhardwaj: We need account for that yes. Do you know how to achieve this?

mnezh: jenkins has secrets for cloud-qa and managed cloud

Pipfile
```diff
@@ -68,3 +68,7 @@ types-pyyaml = ">=6.0"
 types-requests = ">=2.31"
 types-toml = ">=0.10"
 wheel = "==0.41.0"
+pytest-bdd = "==6.1.1"
+playwright = "==1.40.0"
+hac-playwright = { file = "http://h2o-public-test-data.s3.amazonaws.com/e2e-testing/hac_playwright-1.40.0-py3-none-any.whl" }
```
maxjeblick, line 73: [This ](https://github.com/h2oai/h2o-llmstudio/pull/561/files#diff-230078d672f10d17463a8a6265cad825b790885898256a3365be90685caac58d) is currently blocking using 1.40

Makefile
```diff
@@ -26,11 +32,16 @@ setup: pipenv
 setup-dev: pipenv
 	$(PIPENV) install --verbose --dev --python $(PYTHON_VERSION)
 	- $(PIPENV_PIP) install flash-attn==2.3.6 --no-build-isolation --upgrade
+	$(PIPENV) run playwright install
 
 .PHONY: setup-no-flash
 setup-no-flash: pipenv
 	$(PIPENV) install --verbose --python $(PYTHON_VERSION)
 
+setup-ui: pipenv
```
pascal-pfeiffer, line 41: Why does it need another make command, why is -dev not enough? This should get some coverage in the README.md or any other appropriate docs how the tests are to be used.

Pipfile
```diff
@@ -68,3 +68,7 @@ types-pyyaml = ">=6.0"
 types-requests = ">=2.31"
 types-toml = ">=0.10"
 wheel = "==0.41.0"
+pytest-bdd = "==6.1.1"
+playwright = "==1.40.0"
```
pascal-pfeiffer, line 72: afaik, `playwright` is not needed when you already specify `hac-playwright`. Please confirm.

Makefile
```diff
@@ -84,6 +95,24 @@ test: reports
...
+	-o log_level=$(LOG_LEVEL) \
+	-o log_file=reports/tests_ui.log \
+	tests/ui/test.py 2>&1 | tee reports/tests_ui.log
+
+.PHONY: test-ui-headed
+test-ui-headed: 
+	$(PW_DEBUG) $(PIPENV) run pytest \
+	-vvs \
+	--headed \
+	tests/ui/test.py 2>&1 | tee reports/tests.log
```
maxjeblick, line 114: Can you add 
```
		--video=on \
		--screenshot=on \
		--slowmo=100 \
```
(or what is required to view the tests while they are running) to the local test?

Makefile
pascal-pfeiffer: Probably designed to be run on a remote machine connecting to another instance? Again, should get coverage in docs why it has been designed this way and how it is to be used.

itsmunishbhardwaj: precisely, its designed to be able to run from a remote machine. I'll add documentation coverage

Pipfile
itsmunishbhardwaj: youre right, works without playwright

Pipfile
itsmunishbhardwaj: okay, installation the previous available version 1.38.0 instead

mnezh (approved):

maxjeblick (approved):
Thanks a lot for adding the missing Ui tests!
PR looks go to me, we can address subsequent refinements in upcoming PRs.

PR #564, opened 2024-01-15
allow to bypass config checks

### LINKED ISSUE #555: [CODE IMPROVEMENT] Relax config checks for inference only
I think we should relax some config checks when `epochs=0` in 
https://github.com/h2oai/h2o-llmstudio/blob/main/llm_studio/python_configs/cfg_checks.py

For instance, `float16` without LoRA should be allowed.

Maybe it would be also useful if a user can still decide to start the experiment and overrule the config check,

maxjeblick:
This PR adds the possibility to bypass configuration checks in the UI.
Fixes #555


psinger (approved):
Thanks Max - LGTM!

PR #573, opened 2024-01-18
Add stop streaming button 

### LINKED ISSUE #570: [BUG] Chat interface breaks after config error
### 🐛 Bug

The chat interface breaks and cannot be reset after having a configuration error:
image

This happens after setting e.g.:
image

`ValueError: `temperature` (=0.0) has to be a strictly positive float, otherwise your next token scores will be invalid. If you're looking for greedy decoding strategies, set `do_sample=False`.`

It would be in general good if there would be a button allowing to stop the current output streaming, also for very long output that does not stop appropriately.

maxjeblick:
This PR adds a stop streaming button to the chat window. In addition, it adds a configuration check that prevents #570.

Implementation-wise, a new stopping criteria is passed to the generation function that will stop generation whenever a certain env variable (`"STOP_STREAMING"`) is present. I've tried other approaches that will interact with the streaming thread directly (leave generation function as is), but I encountered some errors/issues there.

Fixes #570

### DIFF
--- a/llm_studio/python_configs/text_causal_language_modeling_config.py
+++ b/llm_studio/python_configs/text_causal_language_modeling_config.py
@@ -502,4 +502,9 @@ class ConfigProblemBase(DefaultConfigProblemBase):
             errors["message"] += [
                 "Please enable do sample if you want to use temperature > 0."
             ]
+        if self.prediction.temperature == 0 and self.prediction.do_sample:
+            errors["title"] += ["Temperature needs to be > 0 for do sample"]
+            errors["message"] += [
+                "Please increase temperature if you want to use do sample."
+            ]
         return errors
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Using CLI for fine-tuning LLMs:
 
 ## What's New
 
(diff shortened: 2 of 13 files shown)


llm_studio/python_configs/text_causal_language_modeling_config.py
```diff
@@ -502,4 +502,9 @@ def check(self) -> Dict[str, List]:
             errors["message"] += [
                 "Please enable do sample if you want to use temperature > 0."
             ]
+        if self.prediction.temperature == 0 and self.prediction.do_sample:
```
psinger, line 505: we might also need the other way around when temp is set and do_sample not

psinger (approved):
Awesome!

llm_studio/python_configs/text_causal_language_modeling_config.py
maxjeblick: Good point, will add it. 
Interestingly, in that case `model.generate` will not fail, but just log a warning (and set T=0).

maxjeblick: Decided to leave it as is for the moment, probably better solution is to replace `do_sample = T > 0` and then set a warning if `top_k=0`. In any case, settings will be adapted automatically in this case.

PR #582, opened 2024-01-27
h2o-wave==1.0.0 and SQLAlchemy upgrade

### LINKED ISSUE #581: [CODE IMPROVEMENT] Update to h2o-wave 1.0.0

H2O-wave 1.0.0 was released. We should test and upgrade.

pascal-pfeiffer:
closes #581 

SQLAlchemy following the new type declaration explained here: https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html


maxjeblick (approved):
~~LGTM, did not find any issues with upgrading wave or using `mapped_column` refactoring.~~
See below

maxjeblick (changes requested):
Found an issue that is not present in current main:
- Go to Edit Dataset -> Continue -> Will jump to upload dataset tab

pascal-pfeiffer:
Thank you. Good catch. I changed the critical args to the new check vs `__wave_submission_name__` 
It is probably safe to change most of them now, and it will make the handler.py easier to read.

pascal-pfeiffer:
yes, I noticed too. Do you want it changed?

pascal-pfeiffer:
should all look as before now:
image

explicitly set all footer items to `justify="start"`

maxjeblick (approved):
Thanks for the fixes; dataset edit works, and I haven't noticed any layout changes w.r.t. current main.

PR #583, opened 2024-01-29
rm unused file

pascal-pfeiffer:
This file is not used by the documentation. 
correct (used) file stays in https://github.com/h2oai/h2o-llmstudio/blob/main/documentation/package-lock.json


sherenem (approved):

sherenem:
Thanks for identifying this!

Issue #523, opened 2023-12-04
[BUG] Tooltips are incorrect for some problem types


maxjeblick:
### 🐛 Bug

Some tooltips display incorrect information, e.g. loss function tooltip always explains losses for causal language modeling.
Either:
- Homogenize tooltips across problem types
- Create separate tooltips per problem type

### To Reproduce

Choose Causal Classification Model and browse through tooltips.

### LLM Studio version

165629a

PR #586, opened 2024-01-29
Fix IPO Loss

maxjeblick:
Fixes IPO loss implementation according to https://github.com/huggingface/trl/commit/29d439a2043edf4455b05cae5a1e2ade69d22794, see also https://huggingface.co/blog/pref-tuning


gaborfodor:
I can not select `Beta` with the slider outside of [0.1, 0.5] range

maxjeblick:
Good point, I increased the slider to `beta in [0, 1]`, 0.5 was set too conservative, probably.

gaborfodor (commented):
Two tiny issues with tooltips in experiment settings:
* `Rejected Answer Column` does not have tooltip
* `Loss function` has the original sft tooltip
image

maxjeblick:
Thanks, well spotted!

I've added a tooltip for rejected answers. For the loss function tooltip, there's already an [open issue](https://github.com/h2oai/h2o-llmstudio/issues/523). I'll look into this issue, tooltips are currently a bit diverging.

gaborfodor (approved):
Yeah, makes sense to fix the other tooltips together. Thanks @maxjeblick LGTM! I have run a few experiments the runs were stable and sometimes even better than DPO.

pascal-pfeiffer (approved):

PR #588, opened 2024-01-29
Fix custom backbone text disappearing

### LINKED ISSUE #585: [BUG] LLm Backbone dissapears upon trigger
### 🐛 Bug

When using a custom llm backbone, the name disappears if a page reload is triggered.

### To Reproduce

1. Go to new experiment
2. Enter a custom backbone
3. trigger an action, e.g. personalize

### LLM Studio version
[8cf32e4](https://github.com/h2oai/h2o-llmstudio/commit/8cf32e4528aca1d80a7517e87edd43b58d432e85)

maxjeblick:
Adds custom backbone name to the list of available backbones to avoid custom backbone being rendered empty. 
If the backbone is changed, the custom backbone will be dropped from the list.

Fixes  #585


pascal-pfeiffer (approved):
Thanks, great fix!

PR #591, opened 2024-01-30
wave 0.26.3

pascal-pfeiffer:
intermediate bugfix upgrade to wave 0.26.3 until we resolve all issues to update to wave 1.0.0 (#581)


maxjeblick (approved):
LGTM, thanks for the update!
I've tested the following:
- Can click all navigation items, redirects to the correct panel
- Home -> experiments and Home -> dataset works
- All tabs within a finished experiment work
- Chat works, also clear history and stop streaming
- Delete experiment works
- Compare experiment works, also showing diff only
- New experiment from old one works
- New experiment from old one -> clicking on View experiment tab works
- All Dataset tabs work
- Dataset edit tab works
- Dataset -> new experiment work
- Dataset -> new experiment -> leave experiment view works
- Delete dataset works
- Delete dataset that is used **does not work** completely. It won't be deleted, but pop up is missing. Also present on current main, so issue already exists.
- upload local dataset works
- save settings works
- setting credentials work

pascal-pfeiffer:
Thanks a lot for the detailed tests, @maxjeblick !

> Delete dataset that is used does not work completely. It won't be deleted, but pop up is missing. Also present on current main, so issue already exists.

Nice find! I'll address when upgrading to 1.0.0 and touching the trigger handlers.

Issue #594, opened 2024-01-30
[BUG] Possible bug in validation perplexity metric for DPO?


tmostak:
### 🐛 Bug

I'm not quite sure this is a bug, but after calibrating my hyperparameters for DPO (after running SFT on the model)  to get where validation loss falls through the training process, I am seeing validation perplexity increase, which I didn't think was possible (i.e validation loss and perplexity should be closely linked).

Relevant hyperparameters are Lora R: 64, Lora Alpha 256, Lora Dropout 0.05, Learning rate 0.00001, batch size 1, epochs 1, Gradient Clip 2, Beta 0.05. The dataset is composed of 620 SQL pairs that the model got wrong after SFT, with the corrected output.

Attaching a screenshot so you can see what I mean. Is this expected?

<img width="1538" alt="Screenshot 2024-01-30 at 11 45 47 AM" src="https://github.com/h2oai/h2o-llmstudio/assets/953210/c89dee1a-a1cb-437d-af44-79582ab52fca">

maxjeblick:
Thanks a lot for posting, interesting findings!

I've checked Perplexity implementation, code-wise I haven't found any issues. 
For Dpo, validation loss and perplexity may exhibit different behavior, as dpo calculation is using
`policy_chosen_logps - policy_rejected_logps - (reference_chosen_logps - reference_rejected_logps)` 
whereas Perplexity is using `chosen_logits`, only. 

It is probably a good idea to add additional train/validation metrics (that will be logged to neptune) such as CE loss to better track the experiment.
I've created [this branch](https://github.com/h2oai/h2o-llmstudio/blob/max/dpo_ce_loss/llm_studio/src/models/text_dpo_modeling_model.py#L174) where `SampleAveragedCrossEntropyLoss` is used as a validation loss and ran some experiments on it. So far, val loss is in sync with Perplexity.
![Screenshot 2024-02-05 at 13-01-52 H2O LLM Studio](https://github.com/h2oai/h2o-llmstudio/assets/24281881/7679aa25-9048-4c35-a63a-0fa54f0d3708)

maxjeblick:
I digged a bit deeper into the loss curves in [this branch](https://github.com/h2oai/h2o-llmstudio/tree/max/log_more_dpo_metrics); cross entropy is logged for both accepted as well as rejected samples; alongside with the corresponding perplexity. 

[...]

PR #610, opened 2024-02-07
Log more dpo metrics

maxjeblick:
This PR logs more metrics for DPO training (in particular cross entropy loss). 

I refactored train.py a bit, it is now also possible to automatically log additional metrics/losses by adding them to the model's output dictionary using `additional_log_` prefix (e.g. `outputs["additional_log_chosen_rewards"] = ...`)

Related to #594


train.py
```diff
@@ -85,12 +84,13 @@ def run_eval(
     Returns:
         Validation loss
     """
-
+    torch.inference_mode(mode=True)
     with torch.no_grad():
```
pascal-pfeiffer, line 88: `no_grad` is probably redundant now, but we can leave it in there. 

I remember we had some issues with inference_mode a while back. Was it a necessary change? Otherwise, i'd stick with the old solution.

pascal-pfeiffer (changes requested):
Thanks a lot @maxjeblick 
Super helpful to have more metrics logged. Ideally, we would also want to show them in the UI, but having them on neptune and in the log files, is already much better.

Please have another look on the path change for the gpt_templates. I believe this slipped into your PR somehow. Otherwise, looks good to me!

llm_studio/python_configs/text_causal_language_modeling_config.py
```diff
@@ -318,8 +318,11 @@ def __post_init__(self):
             ),
             allow_custom=True,
         )
+        prompt_template_directory = os.path.join(
+            os.path.dirname(__file__), "../../prompts"
+        )
         self._possible_values["metric_gpt_template"] = possible_values.String(
-            values=(f.split(".")[0] for f in os.listdir("prompts"))
+            values=(f.split(".")[0] for f in prompt_template_directory)
```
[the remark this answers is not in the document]
maxjeblick: Thanks, will check. Reason for the change is to resolve `prompts` directory w.r.t. `llm_studio/python_configs/text_causal_language_modeling_config.py` to be able to run tests with root directory == directory where test is running (this is the default in PyCharm when a test is started within the code editor).

train.py
maxjeblick: Thanks for the catch, inference mode can probably be removed. 
Slipped in from https://github.com/h2oai/h2o-llmstudio/blob/main/train.py#L102 where it is actually not needed.

pascal-pfeiffer (approved):
Thanks again, lgtm!

PR #617, opened 2024-02-22
update lib

haqishen:
## Check

* [x] docker runtime
* [x] flash-attn

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,12 +26,12 @@ pipenv:
 .PHONY: setup
 setup: pipenv
 	$(PIPENV) install --verbose --python $(PYTHON_VERSION)
-	-$(PIPENV_PIP) install flash-attn==2.3.6 --no-build-isolation --upgrade
+	-$(PIPENV_PIP) install flash-attn==2.5.5 --no-build-isolation --upgrade
 
 .PHONY: setup-dev
 setup-dev: pipenv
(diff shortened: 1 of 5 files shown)


maxjeblick:
`test_generation_is_the_same_as_for_causal_language_modeling` is failing.
I assume that https://github.com/huggingface/transformers/commit/3f60d11a8750992287cd0d1f3dbc9df6ffc34288 is the root of the issue ([this line](https://github.com/huggingface/transformers/blame/main/src/transformers/models/llama/modeling_llama.py#L1075)):

```
causal_mask[..., :mask_length].eq(0.0) * attention_mask[:, None, None, :].eq(0.0)
  File "/home/max/.local/share/JetBrains/Toolbox/apps/pycharm-professional/plugins/python/helpers-pro/pydevd_asyncio/pydevd_asyncio_utils.py", line 114, in _exec_async_code
    result = func()
  File "<input>", line 1, in <module>
RuntimeError: The size of tensor a (64) must match the size of tensor b (136) at non-singleton dimension 3
```

Will try to create a minimal example that reproduces this error and potentially create an issue in transformers library.

maxjeblick:
I've resolved the issue that originated from code changes that required `config.max_position_embeddings = 1024` (before it was 32) in `MaxJeblick/llama2-0b-unit-test`.

maxjeblick (approved):
Thanks a lot for the upgrade!
I've tested with several models (including flash attention), looks good!

PR #627, opened 2024-03-04
DeepSpeed ZeRO2

haqishen:
# EXP w/ 2xA6000

| -- | backbone | dtype | LORA | deepspeed  | RAM usage per GPU | Train Time | Valid Time | GPT metric |
| -------- | -------- | ----- | ---- | --------- | ----------------- | ----------------- | ------------------------ | ---------- |
| exp1 | h2oai/h2ogpt-4096-llama2-7b | bfloat16 | True | NA | 35GB | 24min | 8.5min | 0.4373 |
| exp2 | h2oai/h2ogpt-4096-llama2-7b | bfloat16 | True | **ZeRO2** | 15GB | 23.5min | 8min | 0.4084 |
| exp3 | h2oai/h2ogpt-4096-llama2-7b | bfloat16 | True | **ZeRO3** | 14GB | 81min | 101min | 0.3988 |

zero2 trained model huggingface push&pull also works well.


llm_studio/python_configs/text_causal_language_modeling_config.py
```diff
@@ -350,7 +350,8 @@ class ConfigNLPCausalLMEnvironment(DefaultConfig):
     mixed_precision: bool = True
 
     compile_model: bool = False
-    use_deepspeed: bool = False
+    use_deepspeed: str = "NA"
```
maxjeblick, line 353: I'd change "NA" to "No".

Makefile
```diff
@@ -157,6 +157,18 @@ endif
 		-v `pwd`/output:/workspace/output \
 		gcr.io/vorvan/h2oai/h2o-llmstudio:nightly
 
+.PHONY: docker-stop
+docker-stop:
+	@CONTAINERS=$$(docker ps -a -q --filter ancestor=gcr.io/vorvan/h2oai/h2o-llmstudio:nightly); \
+	if [ -n "$$CONTAINERS" ]; then \
+		docker stop $$CONTAINERS; \
+		docker rm $$CONTAINERS; \
```
maxjeblick, line 165: Maybe remove `docker rm`, or rename the make command to be more explicit that the container will be removed, as well.

maxjeblick (changes requested):
Thanks a lot for the PR, deepspeed2 and 3 work as expected, I'm getting similar results as the ones you shared w.r.t. time/gpu consumption using the default settings.

Regarding the code,  I would like to keep `use_deepspeed: bool = False` flag (instead of reassigning it to a string). Otherwise, new experiment from old experiment (with an experiment created before this PR codebase) fails upon reading the config (redefined datatype for config item).

A possible way to implement it is to have a trigger `use_deepspeed` -> `deepspeed_settings` that switches between deepspeed 2 and deepspeed 3.

haqishen:
> Thanks a lot for the PR, deepspeed2 and 3 work as expected, I'm getting similar results as the ones you shared w.r.t. time/gpu consumption using the default settings.
> 
> Regarding the code, I would like to keep `use_deepspeed: bool = False` flag (instead of reassigning it to a string). Otherwise, new experiment from old experiment (with an experiment created before this PR codebase) fails upon reading the config (redefined datatype for config item).
> 
> A possible way to implement it is to have a trigger `use_deepspeed` -> `deepspeed_settings` that switches between deepspeed 2 and deepspeed 3.

@maxjeblick Thanks for the comment!

I have a questions about WAVE:

 Say we're going to keep `use_deepspeed` as bool and adding a new `deepspeed_method` as str within ['ZeRO2', 'ZeRO3'], then I'll have to make those zero2 hyper-params hidden until `use_deepspeed == Ture` and `deepspeed_method == 'ZeRO2'` at the same time. How to implement this via WAVE? What I face now is even `use_deepspeed==False`, those zero2 params shows up because `deepspeed_method` is `ZeRO2` by default.

maxjeblick:
> What I face now is even use_deepspeed==False, those zero2 params shows up because deepspeed_method is ZeRO2 by default

You are right, the current implementation of dependencies shows deepspeed parameters even if use deepspeed is disabled.
I was able to get the triggering to work by:
- Use 
```
    use_deepspeed: bool = False
    deepspeed_model: str = "ZeRO2"
....

        self._possible_values["deepspeed_model"] = ["ZeRO2", "ZeRO3"]


....
         self._nesting.add(
            [
                "deepspeed_model",
            ],
            [Dependency(key="use_deepspeed", value=True, is_set=True)],
        )
        self._nesting.add(
            [
                "deepspeed_reduce_bucket_size",
            ],
            [
                Dependency(key="use_deepspeed", value=True, is_set=True),
            ],
        )
        self._nesting.add(
            [
                "deepspeed_allgather_bucket_size",
            ],
            [
                Dependency(key="use_deepspeed", value=True, is_set=True),
                Dependency(key="deepspeed_model", value="ZeRO2", is_set=True),
            ],
        )
        self._nesting.add(
            [
                "deepspeed_stage3_prefetch_bucket_size",
                "deepspeed_stage3_param_persistence_threshold",
                # "deepspeed_offload_optimizer",
            ],
            [   Dependency(key="use_deepspeed", value=True, is_set=True),
                Dependency(key="deepspeed_model", value="ZeRO3", is_set=True),
            ],
        )
```


- Use `return all_deps == len(dependencies)` instead of `return all_deps >0 ` here https://github.com/h2oai/h2o-llmstudio/blob/main/llm_studio/app_utils/utils.py#L1089. @psinger I guess it should be fine to change this? That is, only show a configuration item if all dependencies are met.

Makefile
haqishen: Actually we don't need to maintain the container, so I remove the `make docker-stop` only adding `make docker-clean-all`

haqishen:
@maxjeblick Thanks! Your code works well from my side.

maxjeblick (approved):
Thanks for implementing ZeRO2, LGTM!

PR #629, opened 2024-03-08
fix wave routing and upd to h2o wave 1.1.0

pascal-pfeiffer:
- update to wave 1.1.0 (fixes default value of `justify="start"`)
- fix routing issues that have been introduced by upgrading to wave to a new major version. We explicitly now check for `q.args.__wave_submission_name__`


maxjeblick (approved):
Thanks, LGTM

PR #650, opened 2024-04-02
Remove add_prefix_space

psinger:
This PR removes the hidden hardcoded option of `add_prefix_space=False`. It will now always respect the settings of the model used.


pascal-pfeiffer (approved):
Thank you. Looks like a reasonable change to me, to just use the default provided from the base model and tokenizer. 

We might want to add a note somewhere that old model configs may be invalid in future transformers versions as transformers does currently not respect that setting for LlamaTokenizerFast.

PR #660, opened 2024-04-08
Apply chat template for HF

### LINKED ISSUE #551: [FEATURE] Automatically add chat template to tokenizer_config.json
### 🚀 Feature

We can automatically add chat template to tokenizer_config.json when pushing to HF or preparing the model download.

### Motivation

Fully integrated chat template in transformers https://huggingface.co/docs/transformers/chat_templating

haqishen:
close #551


pascal-pfeiffer (changes requested):
Thank you for working on that issue @haqishen .

A few things, I noticed:
- chat template is also applied to seq2seq models.
- chat template is not applied in the downloaded models.
- system prompt should be optional
- it probably doesn't need an extra cfg yaml for the unit tests, you can reuse the existing ones.

haqishen:
Hi @pascal-pfeiffer

Thanks for your comments and sorry for the late.

I have a few questions:

> chat template is not applied in the downloaded models.

Chat template actually is not used for building input text within the project (at least for now), so I don't get the point of this ↑ Do I miss something?

> system prompt should be optional

https://github.com/h2oai/h2o-llmstudio/pull/660/files#diff-002a8aeff9716c32e7b21b06df53b9905cae17ae5f75c398705384b1d9507295R107

It's controlled here, and also there is unit test for not using system prompt here:

https://github.com/h2oai/h2o-llmstudio/pull/660/files#diff-f62f85ce93eeb67e1138700aef5af102db092272242ac76b40d7ad6b9a69e71cR47

Or do I misunderstand your comment?

pascal-pfeiffer:
Thanks for asking, I'll try to clarify below:

> Chat template actually is not used for building input text within the project (at least for now), so I don't get the point of this ↑ Do I miss something?

The downloaded model should be identical to the one pushed to huggingface. There is little benefit of maintaining different versions especially in the critical chat template. So, my point is to align the build process for model zip vs huggingface model repo. 


> https://github.com/h2oai/h2o-llmstudio/pull/660/files#diff-002a8aeff9716c32e7b21b06df53b9905cae17ae5f75c398705384b1d9507295R107
> It's controlled here, and also there is unit test for not using system prompt here:
> https://github.com/h2oai/h2o-llmstudio/pull/660/files#diff-f62f85ce93eeb67e1138700aef5af102db092272242ac76b40d7ad6b9a69e71cR47
> Or do I misunderstand your comment?

1. Train a model with system prompt
2. Your PR will add a required system prompt to the chat template
3. User is not able to use the model without providing a system prompt, and face an error "TemplateError: Conversation roles must alternate system/user/assistant/user/assistant/..."

haqishen:
@pascal-pfeiffer

Thanks for your answer, it make a lot of sense.

The current logic is that if an LLM uses a system prompt during training, it must also use a system prompt during generation. Conversely, if the LLM does not use a system prompt during training, it cannot use one during generation.

Your idea is that whether or not an LLM uses a system prompt during training, there should be the option to choose whether to use a system prompt during generation. Is this understanding correct?

pascal-pfeiffer:
> The current logic is that if an LLM uses a system prompt during training, it must also use a system prompt during generation. Conversely, if the LLM does not use a system prompt during training, it cannot use one during generation

> Your idea is that whether or not an LLM uses a system prompt during training, there should be the option to choose whether to use a system prompt during generation. Is this understanding correct?

1. Train with system prompt -> Inference can be done **either with or without** system prompt. Glimpsing at the code below we even support `""` in system prompts, which removes the whole system part of the prompt. Which means this case can be covered in the training data.
```python
    def get_systems(self, cfg, df):
        if cfg.dataset.system_column != "None":
            if cfg.dataset.system_column not in df.columns:
                logger.warning(
                    f"System column {cfg.dataset.system_column} not found."
                    f"Disabling functionality."
                )
                systems = ["" for _ in range(len(self.prompts))]
            else:
                systems = df[cfg.dataset.system_column].astype(str).tolist()
        else:
            systems = ["" for _ in range(len(self.prompts))]
        return systems
```

2. Train without system prompt -> Inference can only be done **without** system prompt (already the case in your PR)

haqishen:
@pascal-pfeiffer 
Updated, please check!

llm_studio/app_utils/hugging_face_utils.py
```diff
@@ -215,7 +228,8 @@ def publish_model_to_hugging_face(
     repo_id = f"{user_id}/{hf_repo_friendly_name(model_name)}"
 
     # push tokenizer to hub
-    tokenizer.chat_template = get_chat_template(cfg)
+    if cfg.problem_type != "text_sequence_to_sequence_modeling":
+        tokenizer.chat_template = get_chat_template(cfg)
```
pascal-pfeiffer, line 232: I don't see the template being used in classification tasks.

llm_studio/app_utils/sections/experiment.py
```diff
@@ -1636,6 +1637,8 @@ async def experiment_download_model(q: Q):
         # See PreTrainedTokenizerBase.save_pretrained for documentation
         # Safeguard against None return if tokenizer class is
         # not inherited from PreTrainedTokenizerBase
+        if cfg.problem_type != "text_sequence_to_sequence_modeling":
+            tokenizer.chat_template = get_chat_template(cfg)
```
pascal-pfeiffer, line 1662: I don't see the template being used in classification tasks.

tests/src/utils/test_chat_template.py
```diff
@@ -8,10 +8,29 @@
...
+        elif msg["role"] == "assistant":
+            expected += f"{cfg.dataset.text_answer_separator}{msg['content']}"
+            if cfg.dataset.add_eos_token_to_answer:
+                expected += eos_token
+        elif msg["role"] == "system":
+            expected += f"{cfg.dataset.text_system_start}{msg['content']}"
+            if cfg.dataset.add_eos_token_to_system:
+                expected += eos_token
+    expected += cfg.dataset.text_answer_separator
+    return expected.replace("\\n", "\n")
```
pascal-pfeiffer, line 27: I honestly liked the old test better with explicit string.
This function itself looks like it needs a unit test.

pascal-pfeiffer (changes requested):
Thanks for the changes @haqishen 
Only a few nitpicks in the code comments.

pascal-pfeiffer (approved):
Thank you for quickly addressing the suggestions. Looks good to me!

PR #687, opened 2024-04-26
log lora params with logger

### LINKED ISSUE #469: [CODE IMPROVEMENT] log trainable params in the UI/add to model summary

Replace `backbone.print_trainable_parameters()` by an explicit log call to propagate the information to the UI.
In addition, we could think of adding it to the model summary tab.

### Motivation

Log num trainable params in a persistent way. Useful when adjusting LORA settings.

pascal-pfeiffer:
closes #469

image


psinger (approved):
lgtm

PR #699, opened 2024-05-08
add config check for learning rate typing

Quetzalcohuatl:
common error is to set learning rate with e notation. For example 3e-4. Yaml parses it as a string

### DIFF
--- a/llm_studio/python_configs/cfg_checks.py
+++ b/llm_studio/python_configs/cfg_checks.py
@@ -102,4 +102,14 @@ def check_for_common_errors(cfg: DefaultConfigProblemBase) -> dict:
             "Deepspeed does not support single GPU training. "
             "Please select more than one GPU or disable deepspeed."
         ]
+
+    # Check if learning rate is accidentally a string
+    # (for example, setting the LR to 3e-4, the yaml considers that a string)
+    if isinstance(cfg.training.learning_rate, str) or isinstance(cfg.training.differential_learning_rate, str):
+        errors["title"] += ["Learning rate in config must be float; e notation is not supported"]
+        errors["message"] += [
+            "Learning rate in config must be float. "
+            "Please convert the learning rate into a purely decimal number."
+        ]
+
     return errors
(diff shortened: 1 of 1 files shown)


pascal-pfeiffer:
Thank you for your PR @Quetzalcohuatl 
After careful checks, I can assure you that the current implementation is working as intended and scientific notation is of course supported in `.yaml` configs. 

Please make sure that you are using the yaml standard used in PyYAML and write scientific notation with a decimal point
```yaml
learning_rate: 3.0e-4
```
or 
```yaml
learning_rate: 3.e-4
```

to be on the safe side, always write the base as a float and add a sign to the exponend (plus or minus).

pascal-pfeiffer:
But your check can indeed make sense in order to catch a wrongfully formatted learning rate. In that case, please update the error message to reflect the correct usage.

pascal-pfeiffer (changes requested):
Thank you for your contribution, @Quetzalcohuatl !
Could you fix the style errors please.

```sh
./llm_studio/python_configs/cfg_checks.py:107:89: E501 line too long (105 > 88 characters)
./llm_studio/python_configs/cfg_checks.py:108:66: BLK100 Black would make changes.
./llm_studio/python_configs/cfg_checks.py:108:89: E501 line too long (111 > 88 characters)
```

Apart from that, the checks are only applied for UI use, where it shouldn't fail anyway, as these fields are type casted.
Two ways to solve probably:
- apply these checks to CLI, too.
- force type casting of these values

Quetzalcohuatl:
@pascal-pfeiffer Imo if you have the code for type-checking for UI, might as well lift-and-shift it to the CLI use-case. That would solve the error where for one of the hyperparams I typed "1.0" as a float but it wanted an integer as "1"

pascal-pfeiffer:
Yes, that would be great. Though, in current form, the code will never get called.
The integration to CLI needs to be added

psinger:
closing this for now, please re-open accordingly

PR #599, opened 2024-01-31
Fake DPO / KTO

psinger:
This PR adds the simple KTO loss.

It currently requires to build pairs of accepted and random rejected samples outside of LLm Studio.

PR #707, opened 2024-05-16
remove cast to float32 and float16 when using CPU for pushing to HF

### LINKED ISSUE #702: [BUG] HuggingFace export does not preserve bfloat16 weights but converts to float16 silently when using CPU for upload
### 🐛 Bug

Native bfloat16 model fine-tuned with bfloat16 gets pushed to HuggingFace as float16

### To Reproduce

1. Choose a HF model like [Llama-3](https://huggingface.co/meta-llama/Meta-Llama-3-8B) with weights natively as bfloat16
2. Fine-tune it using dtype of bfloat16
3. Export it to HuggingFace
4. Note that the config.json specifies the weights of the fine-tuned model as float16 (not bfloat16) as you'd expect

pascal-pfeiffer:
I think recent changes in 3rd party dependencies make this cast to float32 and float16 obsolete and CPU seems to be able to handle it. 

closes #702


psinger (approved):

pascal-pfeiffer:
I found [this](https://github.com/h2oai/h2o-llmstudio/pull/599/files#r1474515499) as an explanation why the cast to float32/16 was added. 
Wdyt, shall we just remove and see when it breaks or wrap in a try/except block? Ideally, we will add a selection for target precision (or quantization) anyway.

psinger:
> I found [this](https://github.com/h2oai/h2o-llmstudio/pull/599/files#r1474515499) as an explanation why the cast to float32/16 was added. Wdyt, shall we just remove and see when it breaks or wrap in a try/except block? Ideally, we will add a selection for target precision (or quantization) anyway.

yes I think we can see when it breaks

PR #709, opened 2024-05-16
Add DoRA support

### LINKED ISSUE #645: [FEATURE] Add DoRA: Weight-Decomposed Low-Rank Adaptation
### 🚀 Feature

Add support for Weight-Decomposed Low-Rank Adaptation (DoRA).
"DoRA decomposes the pre-trained weight into two components, magnitude and direction, for fine-tuning, specifically employing LoRA for directional updates to efficiently minimize the number of trainable parameters. "

Paper: https://arxiv.org/abs/2402.09353
Implementation in Peft: https://github.com/huggingface/peft/releases/tag/v0.9.0

### Motivation

Authors claim that DoRA is more efficient w.r.t. param count as LoRA and achieves better results throughout all tests.

pascal-pfeiffer:
Adds a new option nested under LoRA to use [DoRA](https://arxiv.org/abs/2402.09353) instead.
I didn't see a big boost from DoRA over LoRA, and it comes at quite a cost of runtime and memory footprint. It works well with grad checkpointing but somehow needs a lot of memory without it compared to LoRA.

- updated quite a few dependencies, too. I tested for usual breaks. Deepspeed might need a few more tests, though.

closes #645


tmostak:
Interestingly testing on this branch I'm finding the same thing... definitely runs slower and requires more memory, but in my case with Llama 3 8B the perplexity results were slightly worse with DoRA than with LoRA with all hyperparams otherwise the same.

Going to test on the 70B model over the coming days and will report back.

psinger (approved):
LGTM - thanks!

PR #712, opened 2024-05-17
fix tooltips and remove parent id for classification tasks

### LINKED ISSUE #523: [BUG] Tooltips are incorrect for some problem types

### LINKED ISSUE #528: [CODE IMPROVEMENT] Remove parent ID in classification tasks

Parent ID is not supported in classification tasks, we can remove that from the UI.

pascal-pfeiffer:
- Add description about classification metrics to "metric" tooltip; closes #523 
- remove parent id and connected settings for classification tasks; closes #528


llm_studio/python_configs/text_causal_classification_modeling_config.py
```diff
@@ -54,11 +56,19 @@ def __post_init__(self):
 
         self._possible_values["num_classes"] = (1, 100, 1)
 
+        self._visibility["system_column"] = -1
+        self._visibility["parent_id_column"] = -1
+        self._visibility["text_system_start"] = -1
+        self._visibility["text_prompt_start"] = -1
+        self._visibility["text_answer_separator"] = -1
+        self._visibility["add_prompt_answer_tokens"] = -1
```
psinger, line 64: why remove all these settings? might still be useful

pascal-pfeiffer: Yeah, it is a tradeoff between simplicity and customizability. 
I'd argue that system prompt can at least be removed as it can fully be replaced by Text prompt start in absense of conversation chaining. So, we might want to reduce to pre prompt and post prompt.

psinger: I would still keep start and end token options for example

pascal-pfeiffer: So, keep `text_prompt_start` and `add_eos_token_to_prompt` ? I would agree with that. 

Adding an EOS after answer makes no sense, and I would still argue in favor of removing the system column. One could still select multiple prompt columns.

psinger: I would rather keep `text_answer_separator`, but also `text_prompt_start  `and `add_eos_token_to_prompt `
removing sys is fine to me

pascal-pfeiffer: Ok, sounds good. Will change it accordingly.

psinger (approved):
lgtm!

PR #717, opened 2024-05-22
Improve dataset import

### LINKED ISSUE #338: [CODE IMPROVEMENT] Show system column field during dataset import

During import, the user can currently only specify prompt and answer column.

### LINKED ISSUE #690: [FEATURE] Add ability to specify dataset problem type during data import
### 🚀 Feature
LLM studio supports multiple problem types (Causal Modeling/ Classification), however, during data import, it expects dataset to be named exactly as the problem type, otherwise proper problem type can not be set.

### Motivation

When importing custom dataset named `train.csv` for classification, it would be great to have the ability to select proper problem type.

pascal-pfeiffer:
- User can now select which problem type the dataset should be used for during import. closes #690 
- Nested options such as DPO specific settings will appear automatically.
- Nested options now include system prompt column. closes #338 
- Updated documentation

https://github.com/h2oai/h2o-llmstudio/assets/1069138/921f8a43-34d6-472f-b8ea-d64f6baddaae


sherenem (approved):
LGTM, thanks!

psinger (approved):
LGTM!

Only odd thing is that system is chosen as the default prompt column in dpo.
Could we change that?

pascal-pfeiffer:
done
image

PR #718, opened 2024-05-22
Tokenizer kwargs

psinger:
This PR adds a new setting `tokenizer_kwargs` which can be used to set arbitrary tokenizer arguments.

It also effectively reverts https://github.com/h2oai/h2o-llmstudio/pull/650

We need the removal of the prefix token to make it on-par with the chat template. Unfortunately, this does still not work out of the box for Llama tokenizers, but should work after https://github.com/huggingface/transformers/pull/30964.

As soon as https://github.com/huggingface/transformers/pull/30650 is merged, we can consider moving fully to chat templates.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,6 +99,14 @@ test: reports
     -o log_cli=true -o log_level=INFO -o log_file=reports/tests.log \
     tests/* 2>&1 | tee reports/tests.log'
 
+.PHONY: test-custom
+test-custom: reports
+	@bash -c 'set -o pipefail; export PYTHONPATH=$(PWD); \
+	$(PIPENV) run pytest -s -v --junitxml=reports/junit.xml \
+	--import-mode importlib \
+    -o log_cli=true -o log_level=INFO -o log_file=reports/tests.log \
+    tests/src/datasets/test_text_causal_language_modeling_ds.py --tb=no'
+
 .PHONY: test-ui
 test-ui: reports setup-ui
 	@bash -c 'set -o pipefail; \
--- a/Makefile
+++ b/Makefile
@@ -99,14 +99,6 @@ test: reports
     -o log_cli=true -o log_level=INFO -o log_file=reports/tests.log \
     tests/* 2>&1 | tee reports/tests.log'
 
-.PHONY: test-custom
-test-custom: reports
-	@bash -c 'set -o pipefail; export PYTHONPATH=$(PWD); \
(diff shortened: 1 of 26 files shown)


Makefile
```diff
@@ -99,6 +99,14 @@ test: reports
...
     tests/* 2>&1 | tee reports/tests.log'
 
+.PHONY: test-custom
+test-custom: reports
+	@bash -c 'set -o pipefail; export PYTHONPATH=$(PWD); \
+	$(PIPENV) run pytest -s -v --junitxml=reports/junit.xml \
+	--import-mode importlib \
+    -o log_cli=true -o log_level=INFO -o log_file=reports/tests.log \
+    tests/src/datasets/test_text_dpo_modeling_ds.py --tb=no'
+
```
pascal-pfeiffer, line 109: Can you explain the purpose for this new make command, please?

psinger: sure, to manually switch just single tests without full log outputs

pascal-pfeiffer (approved):
Thank you @psinger !

Could be a tiny issue when forking old experiments where `use_fast` is not default. But not a big deal. For me didn't break anywhere else.

I think it is becoming more and more clear that a standardized way to build chat conversations needs to be used to prevent false formatting. Let's hope this PR is only needed shortly and we can leverage the assistant mask from transformers currently in development there.

tests/src/datasets/test_text_dpo_modeling_ds.py
```diff
@@ -143,15 +143,15 @@ def test_dataset_label_is_correct(df_with_conversation_chain_ids):
...
-            f"<|prompt|>prompt {idx * 5 + 5} "
+            prompt == f"<|prompt|>prompt {idx * 5 + 1}"
+            f"<|answer|>response {idx * 5 + 1}"
+            f"<|prompt|>prompt {idx * 5 + 2}"
+            f"<|answer|>response {idx * 5 + 2}"
+            f"<|prompt|>prompt {idx * 5 + 3}"
+            f"<|answer|>response {idx * 5 + 3}"
+            f"<|prompt|>prompt {idx * 5 + 4}"
+            f"<|answer|>response {idx * 5 + 4}"
+            f"<|prompt|>prompt {idx * 5 + 5}"
```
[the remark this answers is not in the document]
psinger: we can do that as soon as we implement that also in llms, not trusting the chat template fully yet and prefer to manually check for it

PR #726, opened 2024-05-28
Allow to use DPO without LoRA

psinger:
This PR allows to use DPO without LoRA.

It will require more memory and also keeps a copy of the model in GPU memory.

Additional changes:
- Small fix for getting the tokenizer class
- Fixing `use_reentrant=False` for GC as recommended: https://pytorch.org/docs/stable/checkpoint.html#torch.utils.checkpoint.checkpoint
@pascal-pfeiffer please try re-running your exploding lora dpo runs with this pr and see if there is any change in behavior
This setting is also required for dpo setup without LoRA
- Minor fix for mt-bench data to have a potential rejected prompt column, otherwise fails


pascal-pfeiffer (approved):
Thank you, looks good to me!

Issue #727, opened 2024-05-28
[FEATURE] Freezing layers


psinger:
### 🚀 Feature

Add a functionality to allow freezing certain layers of a model.

PR #731, opened 2024-05-29
Option to freeze (non lora) and unfreeze (lora) layers

psinger:
Closes https://github.com/h2oai/h2o-llmstudio/issues/727

This PR addresses the following:

- Removes `force_embedding_gradients` option
- Adds new option `lora_unfreeze_layers` which allows to unfreeze arbitrary backbone layers when using LoRA, such as embedding or head layers
- Adds new option `freeze_layers` which allows to freeze arbitrary backbone layers when using full training (no LoRA)
- Changed default models to `h2o-danube2`
- Removes `enable_input_require_grads` which we manually ran when using gradient checkpointing with kbit lora training, this should not be needed anylonger with `use_reentrant=False`
- Removed restriction to not allow custom options for `combobox` - I had to add a bit of code, but could not find any major issues. Please maybe test a few cases there.


llm_studio/src/utils/modeling_utils.py
```diff
@@ -840,6 +839,13 @@ def create_nlp_backbone(cfg, model_class=AutoModel) -> Any:
                     "likely lead to unstable training without adapters."
                 )
 
+    for name, param in backbone.named_parameters():
+        # freeze base model's layers
+        if any(freeze_layer in name for freeze_layer in cfg.training.freeze_layers):
+            if cfg.environment._local_rank == 0:
+                logger.info(f"Freezing layer: {name}")
+            param.requires_grad = False
```
pascal-pfeiffer, line 847: Let's catch the cases here where LoRA is enabled and print a warning that freezing is ignored.

llm_studio/src/utils/modeling_utils.py
```diff
@@ -1029,10 +1035,22 @@ def prepare_lora(cfg, backbone):
...
 
+    for name, param in backbone.named_parameters():
+        # unfreeze base model's layers
+        if any(
+            unfreeze_layer in name
+            for unfreeze_layer in cfg.training.lora_unfreeze_layers
+        ):
+            if cfg.environment._local_rank == 0:
+                logger.info(f"Unfreezing layer: {name}")
+            param.requires_grad = True
```
pascal-pfeiffer, line 1052: Same here, catch cases where LoRA is disabled.

pascal-pfeiffer (changes requested):
Very neat feature, thank you!

We need to make it a bit more robust vs. contradicting settings. E.g. I can set LoRA to false, freeze layers, then set LoRA true again and run an experiment. This will still cause the freezing loop to be called. Especially for CLI usage, we should catch these and print a warning. 

```
Lora: True
2024-05-30 06:53:30,276 - INFO: Freezing layer: model.layers.0.self_attn.q_proj.weight
2024-05-30 06:53:30,276 - INFO: Freezing layer: model.layers.0.self_attn.k_proj.weight
2024-05-30 06:53:30,276 - INFO: Freezing layer: model.layers.0.self_attn.v_proj.weight
2024-05-30 06:53:30,276 - INFO: Freezing layer: model.layers.0.self_attn.o_proj.weight
```

Is there any way to reset these UI components when they are hidden in the UI?

llm_studio/src/utils/modeling_utils.py
psinger: that one cant happen because it is in the lora path

llm_studio/src/utils/modeling_utils.py
psinger: I will just move it into the correct if without warning as this setting should anyways only be used without lora and docs should be clear there.

pascal-pfeiffer (approved):
Thank you for the change, lgtm!

Quetzalcohuatl:
Changing use_reentrant from True to False breaks deepspeed zero3 with gradient checkpointing. See github.com/huggingface/transformers/issues/34928

Issue #734, opened 2024-05-30
[DOCS] Duplicate Questions in the FAQ's


meganjkurka:
The FAQ has two different answers to how much data do you need to fine tune:

* https://docs.h2o.ai/h2o-llmstudio/faqs#how-much-data-is-generally-required-to-fine-tune-a-model
* https://docs.h2o.ai/h2o-llmstudio/faqs#how-many-records-are-recommended-for-fine-tuning

PR #739, opened 2024-06-04
Improve FAQ page

sherenem:
This PR:
- removes a duplicate question and answer (resolves https://github.com/h2oai/h2o-llmstudio/issues/734)
- adds an FAQ about the max dataset size that LLM Studio can handle


pascal-pfeiffer (approved):
Thank you, lgtm!

PR #741, opened 2024-06-04
Max Length rework

psinger:
This PR adresses several related points:

- It removes the settings `max_length_prompt` and `max_length_answer` and only keeps `max_length`. It will always use that setting to truncate the left side of the whole input.
- It will first truncate all prompts on the left side, and all answers on the right side with `max_length`. 
- It removes the strict requirement for the system prompt to always be on the left side of the input, even if truncated. It will also be truncated if the whole input exceeds max length.
- Removes the strict forced requirement for the last label to always be eos token, if setting is set. 
- It is now passing correct `position_ids` to the forward method of the backbone.


llm_studio/python_configs/text_causal_language_modeling_config.py
```diff
@@ -174,8 +174,8 @@ def __post_init__(self):
         )
         self._possible_values["differential_learning_rate_layers"] = (
             possible_values.String(
-                values=("backbone", "embed"),
-                allow_custom=False,
+                values=("backbone", "embed", "head"),
```
pascal-pfeiffer, line 177: Why adding `head` to causal language modeling? Isn't it called `lm_head` for most models?

psinger: It is just a substring match, so `head` works for most/all backbones.

pascal-pfeiffer (approved):
Thank you for the quick changes, looks good to me!

PR #742, opened 2024-06-04
[FEATURE] Grid Search for Hyperparameters

### LINKED ISSUE #251: [FEATURE] Support for grid search over hyperparameters
### 🚀 Feature

Add the capability to the UI to kick off a grid-search over a set of hyperparameters (with specified search increments for continuous parameters, and specified attributes for categorical parameters), which would search for the combination of hyperparameters that yielded the best validation result.

### Motivation

Currently attempting to find the best hyperparameters for model training requires manual launching of many experiments testing the various hyperparameter settings. It would be useful if the UX supported kicking off a "meta-experiment" which would systematically sweep a specified set of hyperparameters, allowing the user to specify the step increment for continuous parameters and allowed categorical attributes (i.e. model backbone) for categorical. 

I would assume simple grid search would be a good start, but if possible the hyperparameter search feature could support other search algorithms as desired.

To save space, I assume that the user should be able to specify that the sub-models not actually be saved after training, or perhaps only the best model be saved.

Ideally the different experiments could be grouped into a single meta-experiment for easy comparison of the impact of hyperparameter changes on model accuracy, but if that's difficult they could just be output as individual experiments (would just take more work on the part of the user to manually sift through the results, but at least they wouldn't have to orchestrate the process of kicking off all the experiments).

Also ideally:

1) Parameter steps could be specified as linear increments or as increments by a power, i.e. LORA rank could be incremented by powers of two.

2. Certain hyperparameters could be specified as dependencies on hyperparameters being searched, i.e. LORA alpha could be specified as always 2X the rank.

I know this is probably a massive feature and may not fit with the goals of this project, but I think it would provide a lot of power for building the best models possible in the most automated way possible.

### LINKED ISSUE #738: [BUG] Default model is added twice to the LLM Backbone list
### 🐛 Bug

image

### To Reproduce

Click "Create experiment"

### LLM Studio version
introduced in 6ab4572a3c0b3871490c69d41ec85e6f281e0b34

pascal-pfeiffer:
This PR adds a new major feature for tuning: Hyperparameter Grid Search
When enabled on the top of a "Create experiment" screen, it enables to select a grid of settings that are all added to the experiment queue with a single click. The "Run experiment" button now shows the number of experiments that will be started with the grid search feature.
closes #251

Minor additional changes
- fix #738 
- Adds `parent_id` as default
- Upgrade dependencies, (eps was removed from `log_loss` in sklearn 1.5.0)
- return val_metric from run() to prepare for optimization

image

image

image


tmostak:
Awesome to see this, can't wait to try it out!

pascal-pfeiffer:
Due to a potential bug in wave (https://github.com/h2oai/wave/issues/2342), I changed the display for the number of experiments to a dialog. This also reduces the number of required trigger reloads, which is especially helpful in remote use of H2O LLM Studio.

image

psinger:
@pascal-pfeiffer 

Is there any "easy" way to consider the changed values when switching back and forth between grid search / non grid search?

So for example, I change `max_length` and then decide to switch to grid search, all settings are reset. Same thing the other way around.

Or maybe even have separate buttons for that instead of the toggle?

psinger:
Somehow it says "Starting 12 experiments" although it starts only 9 (which is correct).

psinger:
But I just selected a grid of 9 experiments, so not sure how it comes up with 12, maybe some hidden value somewhere.

psinger (approved):
Amazing functionality @pascal-pfeiffer 

All works well, let's merge as-is to get the functionality out and fix smaller issues along the way.

Issue #732, opened 2024-05-29
[FEATURE] Implement SimPO


psinger:
### 🚀 Feature

Add https://arxiv.org/abs/2405.14734 for DPO problem type

PR #749, opened 2024-06-06
CPO & SimPO Loss

psinger:
This PR implements new CPOLoss and SimPOLoss

Please review after https://github.com/h2oai/h2o-llmstudio/pull/753

Closes https://github.com/h2oai/h2o-llmstudio/issues/732


llm_studio/python_configs/text_dpo_modeling_config.py
```diff
@@ -74,7 +75,21 @@ class ConfigDPOTraining(ConfigNLPCausalLMTraining):
...
+        self._grid_search_values["beta"] = (0.1, 0.15, 0.20, 0.25, 0.4, 0.5)
+        self._grid_search_values["simpo_gamma"] = (0.5, 0.75, 1, 1.25, 1.5, 1.75, 2)
+
+        self._grid_search_iscustom["beta"] = True
+        self._grid_search_iscustom["simpo_gamma"] = True
+
+        self._nesting.add(
+            ["simpo_gamma"],
+            [Dependency(key="loss_function", value="SimPOLoss", is_set=True)],
+        )
```
pascal-pfeiffer, line 89: That was something that I wanted to avoid when adding the grid search. Now, when starting a grid search with DPO + SPO and selecting SPO only grid search params, the warning message will return a wrong number of experiments being started. 
When actually starting the experiment, this is catched though, and duplicate experiments are not started. 

Maybe this is fine and kind of an edge case, but would be great at least open an issue for it, if you decide not to touch that in this PR and want to keep this a Grid search value.

pascal-pfeiffer (commented):
Thank you for the integration @psinger !

It seems to work nicely, though default values give bad results.
We might want to consider not to rename the Loss functions to avoid issues when starting from older experiments. 
Also, there is a minor issue with the calculation of the number of experiments being started in grid search.

llm_studio/src/losses/text_dpo_modeling_losses.py
```diff
@@ -140,9 +201,11 @@ class Losses:
 
     _losses = {
         "DPOLoss": DPOLoss,
-        "HingeLoss": HingeLoss,
-        "IPOLoss": IPOLoss,
+        "DPOHingeLoss": DPOHingeLoss,
+        "DPOIPOLoss": DPOIPOLoss,
```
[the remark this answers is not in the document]
psinger: I doubt that is a big issue. The reason for re-naming is that we might want to add other options there, and then we need to differentiate between DPO and others.

psinger:
> It seems to work nicely, though default values give bad results.

Bad in what way? It is based on best experiments I got for them, but it might be more sensitive depending on what model / data one uses.

llm_studio/python_configs/text_dpo_modeling_config.py
psinger: What exactly do you mean with avoiding, you mean the dependency?

llm_studio/src/losses/text_dpo_modeling_losses.py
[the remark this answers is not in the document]
psinger: Yes, correct. I thought about renaming it to "Preference Modeling" but decided against it for now.

llm_studio/python_configs/text_dpo_modeling_config.py
pascal-pfeiffer: I avoided allowing grid search on nested structures, yes. We'll need to add the filtering to the experiment number calculation, too.

pascal-pfeiffer:
> > It seems to work nicely, though default values give bad results.
> 
> Bad in what way? It is based on best experiments I got for them, but it might be more sensitive depending on what model / data one uses.

Just running the default experiment with Loss function changed to SPO gives for example these predictions: 
image

But it is probably fine, it is expected that these things need tuning.

pascal-pfeiffer (approved):
Thank you @psinger , lgtm!

PR #753, opened 2024-06-07
Allow to mask all but last answer

psinger:


llm_studio/python_configs/text_dpo_modeling_config.py
```diff
@@ -27,7 +27,6 @@ class ConfigDPODataset(ConfigNLPCausalLMDataset):
     # Always have full chat history.
     # Chosen/Rejected prompt are only at the end of a conversation.
     limit_chained_samples: bool = True
-    mask_prompt_labels: bool = True
```
pascal-pfeiffer, line 30: I'd still leave it there, as this this explicitly needs to be True for DPO. Even if we would change defaults in the Base Class, it should be overwritten here.

llm_studio/python_configs/text_dpo_modeling_config.py
```diff
@@ -27,7 +27,6 @@ class ConfigDPODataset(ConfigNLPCausalLMDataset):
     # Always have full chat history.
     # Chosen/Rejected prompt are only at the end of a conversation.
     limit_chained_samples: bool = True
```
pascal-pfeiffer, line 29: With the new setting, would it be possible to make this customizable again?
Say, only last answers or all answers (but only chained up to the one in question)

pascal-pfeiffer (approved):
LGTM!
Left a minor suggestion.

For CLI / yaml users, the options might be getting a bit tricky now. Would also be nice to check for errors there. And would be good to have the setting added to the example.yaml

llm_studio/python_configs/text_dpo_modeling_config.py
psinger: not sure I fully get it, could you please elaborate?

pascal-pfeiffer: Realized that the PR solves it already. In the end there are two options as in standard Causal Language Modeling:
- calculate loss on all answers 
- calculate loss on just the last answer

PR #771, opened 2024-07-11
H2O drive support for dataset import

pascal-pfeiffer:
This PR adds a new feature
- Importing datasets with [h2o-drive](https://pypi.org/project/h2o-drive/)

Minor changes:
- Shows import errors in the UI in addition to the logs
- Update Docker base image to ubuntu22
- Fixed CVEs from base image and added a scan


llm_studio/app_utils/sections/dataset.py
```diff
@@ -324,7 +354,7 @@ async def dataset_import(
                 ui.textbox(
                     name="dataset/import/kaggle_command",
                     label="Kaggle API command",
-                    value=default_cfg.kaggle_command,
+                    placeholder="kaggle competitions download -c dataset",
```
psinger, line 357: the placehold will not be used though, or will it?
user needs to enter something there then

Issue #757, opened 2024-06-18
H2O LLM Studio error "Unknown exception" when running app


Quickedu:
I had this error just after I started one experiment. When it happened first time, it crashed the whole thing. I have restarted H2O and tried to reinstall and get GitHub directories but it does not work. Now, when I run H2O, in screen Home, View Experiment and Create experiment, I see the same screen message (I upload screenshot of it).

This is the report H2O has given:

Report Issue: H2O LLM Studio at http://192.168.1.101:10101/

q.app
script_sources: ['/_f/60417ac8-9c68-4bed-8f1c-f875e52f45ef/tmpftsu6rqn.min.js']
initialized: True
version: 1.7.0-dev
name: H2O LLM Studio
heap_mode: False
wave_utils_stack_trace_str: ### stacktrace
Traceback (most recent call last):

  File "/home/midue/h2o-llmstudio/./llm_studio/app_utils/handlers.py", line 78, in handle
    await home(q)

  File "/home/midue/h2o-llmstudio/./llm_studio/app_utils/sections/home.py", line 21, in home
    experiments = get_experiments(q)

  File "/home/midue/h2o-llmstudio/./llm_studio/app_utils/utils.py", line 1626, in get_experiments
    info = get_experiments_info(df, q)

  File "/home/midue/h2o-llmstudio/./llm_studio/app_utils/utils.py", line 1449, in get_experiments_info
    with SqliteDict(f"{row.path}/charts.db") as logs:

  File "/home/midue/llmstudio/lib/python3.10/site-packages/sqlitedict.py", line 157, in __init__
    raise RuntimeError('Error! The directory does not exist, %s' % dirname)

RuntimeError: Error! The directory does not exist, /home/midue/h2o-llmstudio/output/user/B.1.4.1.2_Casual-Modeling_h2oai-h2o-danube2-i.8b-base_Val.Size0.2_832-832-1664_bfloat16_LRate0.01_DiffLRate0.001_WarmEpoch1_Batch12_Epoch100_BLEU_AllDropout0.2_NumBeams-3_.1

q.user
q.client
app_db: <llm_studio.app_utils.db.Database object at 0x74b06815bfd0>
client_initialized: True
mode_curr: error
theme_dark: True
credential_saver: .env File
default_aws_bucket_name: bucket_name
default_azure_conn_string: 
default_azure_container: 
default_kaggle_username: 
set_max_epochs: 1000
set_max_batch_size: 256
[...]

PR #776, opened 2024-07-19
Do not break app when experiment folder is missing

psinger:
Do not break the whole app as soon as an experiment folder is missing. 

Closes
https://github.com/h2oai/h2o-llmstudio/issues/757

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,10 +2,9 @@
 input/
 notebooks/
 demo_data/
-output/
-output_old/
+output*/
 tmp/
-data/
+data*/
 examples/data_oasst2
 examples/output_oasst2
 data_old/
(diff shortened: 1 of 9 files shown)


pascal-pfeiffer (commented):
Thank you! LGTM

Can/should we add this at least to the logger when the experiment was deleted? Ideally, The user would be notified and can remove these from the database.

pascal-pfeiffer (changes requested):
I would really ask the user and delete the corresponding entries in the database. Current behavior is, that the experiment is still shown and the app subsequently crashes when I open the experiment and click on any of the other tabs, such as summary.

I mean, it is certainly better than having the app in a fully broken state, so we can also add in in another PR and I am happy to approve this as the first step.

psinger:
I find it personally this way better. Just keep it as-is. It is fine that the tabs fail, but then the whole app does not fail like it did now. It is a very edge case that this happens anyways.

pascal-pfeiffer (approved):

PR #800, opened 2024-08-01
H2O Danube3 and update to default model list

pascal-pfeiffer:
New proposal for default models in H2O LLM Studio. Replacing some of the older ones with new releases.


psinger (approved):
LGTM

But 4B might be too large as default?

Issue #755, opened 2024-06-11
[FEATURE] Export Chat session


psinger:
### 🚀 Feature

Add functionality to export current chat session / history and share it with others.

PR #804, opened 2024-08-02
Copy chat to clipboard

psinger:
New button "Copy to clipboard" to copy the current chat session to the clipboard.

Closes https://github.com/h2oai/h2o-llmstudio/issues/755


llm_studio/app_utils/sections/chat_update.py
```diff
@@ -88,6 +88,20 @@ async def update_chat_window(q):
     q.page["experiment/display/chat"].data[-1] = message
 
 
+async def chat_copy(q: Q) -> None:
+
+    chat_messages = [
+        f"{'USER' if t[1] == USER else 'BOT'}: {t[0]}"
```
pascal-pfeiffer, line 94: ASSISTANT is probably more common

```suggestion
        f"{'USER' if t[1] == USER else 'ASSISTANT'}: {t[0]}"
```

PR #808, opened 2024-08-06
run nvidia-smi at startup

### LINKED ISSUE #807: [FEATURE] Add an early `nvidia-smi` call at the startup routine to expose GPU issues
### 🚀 Feature

Add an early `nvidia-smi` call at the startup routine to expose GPU issues.

### Motivation

Expose issues with GPU(s) early on without needing to interact with the UI

pascal-pfeiffer:
closes #807

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -225,6 +225,7 @@ endif
 		--shm-size=64g \
 		--init \
 		--rm \
+		--it \
 		-u `id -u`:`id -g` \
 		-p 10101:10101 \
 		-v `pwd`/llmstudio_mnt:/home/llmstudio/mount \
--- a/Makefile
+++ b/Makefile
@@ -180,6 +180,8 @@ test-ui-github-actions: reports setup-ui
 .PHONY: wave
 wave:
 	nvidia-smi && \
+	H2O_WAVE_APP_ACCESS_KEY_ID=dev \
+	H2O_WAVE_APP_ACCESS_KEY_SECRET=dev \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -179,13 +179,17 @@ test-ui-github-actions: reports setup-ui
...
+	H2O_WAVE_APP_ACCESS_KEY_ID=dev \
+	H2O_WAVE_APP_ACCESS_KEY_SECRET=dev \
 	H2O_WAVE_MAX_REQUEST_SIZE=25MB \
 	H2O_WAVE_NO_LOG=true \
 	H2O_WAVE_PRIVATE_DIR="/download/@$(WORKDIR)/output/download" \
 	$(PIPENV) run wave run app
 
 .PHONY: llmstudio
 llmstudio:
+	nvidia-smi && \
```
psinger, line 192: would it be enough to keep it only for `make llmstudio`?

pascal-pfeiffer: Yes, indeed not really needed for dev. Good point, will change it.

psinger (approved):

Makefile
psinger: Want to keep it for `llmstudio-conda`?

pascal-pfeiffer: yes, it seems that this isn't meant for dev, as it has the `--no-reload` flag?

psinger: fair enough

PR #813, opened 2024-08-07
chore: update experiment.py

eltociear:
minor fix


psinger:
thanks, will fix this separately

Issue #805, opened 2024-08-05
[FEATURE] Multilabel classification


psinger:
### 🚀 Feature

Support multilabel causal classification

PR #815, opened 2024-08-07
Classification & Regression multilabel

psinger:
In this PR we add multi-target support for both classification and regression.

The following code-adaptation have been made:

- For classification and regression the answer column can now be a selection of multiple targets, changing the setting to a tuple
- Multi label will only work with BCE
- For plotting and visualizations, the target columns are concatenated as string
- The same applies for predictions
- Classification predictions are now consistently post-process in `postprocess_output` instead of in the individual metrics
- The validation csv file now contains the probabilities instead of the hard predictions
- The validation pickle file now contains `logits`, `probabilities` and `predictions`
- For regression, we set the regression head size by number of answer columns
- For classification, the `num_classes` needs to still be set. Potentially would be easier to also start deriving that automatically, would also allow to get rid of all the error checks
- Added metric tests for regression and adjusted for classification
- Adjusted integration tests

When reviewing, main potential for bugs is with respect to post processing, loss and metrics.

Closes https://github.com/h2oai/h2o-llmstudio/issues/805


llm_studio/src/datasets/text_causal_regression_ds.py
```diff
@@ -28,11 +28,28 @@ def __getitem__(self, idx: int) -> Dict:
...
 
     def postprocess_output(self, cfg, df: pd.DataFrame, output: Dict) -> Dict:
-        output["logits"] = output["logits"].float()
-        preds = output["logits"]
-        preds = np.array(preds).astype(float).astype(str).reshape(-1)
+        output["predictions"] = output["predictions"].float()
+        preds = []
+        for col in np.arange(len(cfg.dataset.answer_column)):
+            preds.append(
+                np.round(output["predictions"][:, col].cpu().numpy(), 3).astype(str)
```
pascal-pfeiffer, line 35: specific reason for the rounding here?

psinger: yeah this is shown in the visualizations and with lots of digits not great to read

pascal-pfeiffer: Then, why not truncate in the visualization? The current implementation also rounds the downloadable predictions.

llm_studio/src/utils/data_utils.py
```diff
@@ -149,6 +149,7 @@ def read_dataframe_drop_missing_labels(
     df[input_cols] = df[input_cols].fillna("").astype(str)
     if (
         hasattr(cfg.dataset, "answer_column")
+        and not isinstance(cfg.dataset.answer_column, list)
```
pascal-pfeiffer, line 152: Why list and not tuple? I think we need to be more strict with the types here so be sure we don't miss a case

psinger: because we cast it to list at some point, we can make it a check for tuple and list to make sure

llm_studio/src/datasets/text_causal_regression_ds.py
psinger: It doesnt because `output["predicted_text"]` is not used there

pascal-pfeiffer: Ah, right, that was changed in the PR. Still is odd for the exported dataframe to have rounded values

psinger: Hmh, debatable. Would be tricky to add something different there. The pickle should be used anyways for exact values.

tests/integration/test_causal_binary_classification_modeling_cfg.yaml
```diff
@@ -12,7 +12,8 @@ dataset:
     add_eos_token_to_answer: false
     add_eos_token_to_prompt: false
     add_eos_token_to_system: false
-    answer_column: binary_label
+    answer_column: 
+    - binary_label
```
pascal-pfeiffer, line 16: This required change is a bit unfortunate as the error message is rather cryptic for CLI users when using previously well working yamls.

```
  File "/home/xxx/h2o-llmstudio/train.py", line 722, in <module>
    run(cfg=cfg)
  File "/home/xxx/h2o-llmstudio/train.py", line 530, in run
    train_dataset = get_train_dataset(train_df=train_df, cfg=cfg)
  File "/home/xxx/h2o-llmstudio/llm_studio/src/utils/data_utils.py", line 396, in get_train_dataset
    train_dataset: Dataset = cfg.dataset.dataset_class(
  File "/home/xxx/h2o-llmstudio/llm_studio/src/datasets/text_causal_classification_ds.py", line 19, in __init__
    check_for_non_int_answers(cfg, df)
  File "/home/xxx/h2o-llmstudio/llm_studio/src/datasets/text_causal_classification_ds.py", line 104, in check_for_non_int_answers
    x for x in df[column].values if not is_castable_to_int(x)
  File "/home/xxx/.local/share/virtualenvs/h2o-llmstudio-tT3gHl3a/lib/python3.10/site-packages/pandas/core/frame.py", line 4102, in __getitem__
    indexer = self.columns.get_loc(key)
  File "/home/xxx/.local/share/virtualenvs/h2o-llmstudio-tT3gHl3a/lib/python3.10/site-packages/pandas/core/indexes/base.py", line 3812, in get_loc
    raise KeyError(key) from err
KeyError: 'w'
```

documentation/docs/tooltips/experiments/_answer-column.mdx
```diff
@@ -1,3 +1,5 @@
 The column in the dataset containing the expected output.
 
-For classification, this needs to be an integer column starting from zero containing the class label, while for regression, it needs to be a float column.
\ No newline at end of file
+For classification, this needs to be an integer column starting from zero containing the class label, while for regression, it needs to be a float column.
+
+Multiple target columns can be selected for classification and regression.
```
pascal-pfeiffer, line 5: This might need some examples to illustrate.

tests/integration/test_causal_regression_modeling_cfg.yaml
```diff
@@ -12,7 +12,9 @@ dataset:
     add_eos_token_to_answer: false
     add_eos_token_to_prompt: false
     add_eos_token_to_system: false
-    answer_column: regression_label
+    answer_column: 
+    - regression_label
+    - regression_label
```
pascal-pfeiffer, line 17: This is something not possible in the UI, I'd suggest to update the dataframe to have multiple regression label columns. Adding one column twice should probably even result in an error.

pascal-pfeiffer (changes requested):
All looks very clean and works well for the usual cases, thank you!

I think we might want to improve a bit on the documentation and on error handling. Especially, for CLI users.

The required change to use a list/tuple format for the answer column is a bit unfortunate as the error message is rather cryptic for CLI users when using previously well working yamls.

```
  File "/home/xxx/h2o-llmstudio/train.py", line 722, in <module>
    run(cfg=cfg)
  File "/home/xxx/h2o-llmstudio/train.py", line 530, in run
    train_dataset = get_train_dataset(train_df=train_df, cfg=cfg)
  File "/home/xxx/h2o-llmstudio/llm_studio/src/utils/data_utils.py", line 396, in get_train_dataset
    train_dataset: Dataset = cfg.dataset.dataset_class(
  File "/home/xxx/h2o-llmstudio/llm_studio/src/datasets/text_causal_classification_ds.py", line 19, in __init__
    check_for_non_int_answers(cfg, df)
  File "/home/xxx/h2o-llmstudio/llm_studio/src/datasets/text_causal_classification_ds.py", line 104, in check_for_non_int_answers
    x for x in df[column].values if not is_castable_to_int(x)
  File "/home/xxx/.local/share/virtualenvs/h2o-llmstudio-tT3gHl3a/lib/python3.10/site-packages/pandas/core/frame.py", line 4102, in __getitem__
    indexer = self.columns.get_loc(key)
  File "/home/xxx/.local/share/virtualenvs/h2o-llmstudio-tT3gHl3a/lib/python3.10/site-packages/pandas/core/indexes/base.py", line 3812, in get_loc
    raise KeyError(key) from err
KeyError: 'w'
```

We should handle that better as it is an expected source of error. For the other improvements on documentation, we can handle it in subsequent iterations.

pascal-pfeiffer (approved):
Thank you for the quick changes, lgtm!

PR #817, opened 2024-08-08
Refactor local logging

### LINKED ISSUE #584: [CODE IMPROVEMENT] Logger code improvement

Current code is convoluted by `if rank==0` statements for logging. We should write a wrapper for that.

pascal-pfeiffer:
- Move the local_rank == 0 check into the logger initialization to simplify code
- Adds an option to log all ranks for an experiment.

closes #584


psinger (approved):
Thanks, that makes it a lot cleaner!

For `Log All Ranks` I would personally suggest to make this a global setting instead of an experiment setting.

Theoretically we could also add the same functionality for `cfg.logging._logger.log` to also get rid of the if switches there.

pascal-pfeiffer:
> For Log All Ranks I would personally suggest to make this a global setting instead of an experiment setting.

If was going for it initially, but would need updating and passing of an environment variable to `train.py`. I would suggest that we add it as a global setting for the default value. I can do it in another PR as it perfectly builds upon this one.

> Theoretically we could also add the same functionality for cfg.logging._logger.log to also get rid of the if switches there.

Yes, though, most only support single rank logging there, so, I'd suggest we hardcode this in the external logger initialization and don't make it configuarable.

Issue #700, opened 2024-05-08
[FEATURE] Option to plot train/eval plots with epoch instead of step on x-axis


tmostak:
### 🚀 Feature

Allow epoch to be optionally used as the x-axis of training/eval charts for easier comparison between runs with different amounts of training pairs.

### Motivation

I'm often experimenting with adding or removing subsets of training data, either at dataset creation time by adding or deleting categories of prompts, or via the sampling slider in the H2O LLM Studio UX. Currently, all training and eval loss and perplexity graphs (and any other metric graphs) use step as the x-axis, but it would be easier to compare training runs with different numbers of training pairs if you could switch the x-axis to epoch, that way two training runs X% of the way finished would line up. See the image below for an example of how it is hard to compare two training runs with dissimilar numbers of training pairs (here for the shorter run I removed various prompt templates teaching specific skills/tasks from the training dataset, as I want to see if that improves performance on one specific task I care about and which solely makes up the eval set).

<img width="1670" alt="Screenshot 2024-05-08 at 2 14 21 PM" src="https://github.com/h2oai/h2o-llmstudio/assets/953210/292caae8-9be3-43fd-ac12-3db1fa108076">

pascal-pfeiffer:
I think we can close this as a solution was merged in https://github.com/h2oai/h2o-llmstudio/pull/820

If not sufficient, please reopen.

PR #820, opened 2024-08-13
New setting: log_step_size

psinger:
Introduces a new setting `log_step_size`.

Specifies the interval for logging during training. Two options are available:

- **Absolute**: The default setting. Uses the total number of training samples processed as the x-axis for logging.
- **Relative**: Uses the proportion of training data seen so far as the x-axis for logging.

The new relative setting will always range from 0 to 1, even in case of multiple epochs. But it solves the proposed change in https://github.com/h2oai/h2o-llmstudio/issues/700 meaning that multiple runs with different amounts of data can be compared relatively.

Additionally, cleaned some `step_size` specifications for internal logging. Please confirm when review if I didnt miss anything there for that to be required.

image


pascal-pfeiffer (approved):
Very handy! All lgtm, thank you.

pascal-pfeiffer:
And let's add the tooltip to the documentatio page, too, please.

PR #824, opened 2024-08-14
Fixes HF token issue

psinger:
Fixes a regression for providing HF token when downloading model.

Also removes custom check for outdated `RWForCausalLM` model.


pascal-pfeiffer (commented):
Thank you for the fix. Can we add an automated test for that?

pascal-pfeiffer (approved):
Approved, but let's consider adding a test for it to catch regressions.

Issue #644, opened 2024-03-21
[BUG] Tokenizer config has add_bos_token=true while LLM Studio is training with add_special_tokens=False


pascal-pfeiffer:
### 🐛 Bug

The generated `tokenizer_config.json` has `add_bos_token=true` while H2O LLM Studio is training with `add_special_tokens=False`.
Using the default AutoTokenizer, this leads to different behaviors. 

We should be explicit/correct about it and set `add_bos_token=false`

### To Reproduce

Fine tune a model and download / push to HF

### LLM Studio version
<=1.4.1, b70b04f68d16ae73524d7f38f45e571ddb92cfc3

psinger:
`add_eos_token=false` as well

PR #827, opened 2024-08-14
Tokenizer config fix

psinger:
Closes https://github.com/h2oai/h2o-llmstudio/issues/644

We can always set this to False as we do not use it and it should not be used downstream.
We add those parts to the chat template.


pascal-pfeiffer (approved):
lgtm

PR #833, opened 2024-08-16
New default datasets for classification and regression

psinger:
Adding two new default datasets for classification and regression


pascal-pfeiffer (approved):
Thank you! Working great :)

PR #834, opened 2024-08-18
Raise Error if missing system col in train DataFrame

### LINKED ISSUE #600: [CODE IMPROVEMENT] if user specifies a “system” column but it doesnt exist, it should error out instead of continue running silently

if system column not in train dataframe.coljmns or in valid columns, then error out

### Motivation

Otherwise user might erroneously believe they are using a system column

pascal-pfeiffer:
I was initially going for DataFrame checks in `check_config_for_errors`, but that overcomplicates the code in a way that I don't feel is necessary for this single check. So, added to `train.py` instead.

```
  File "./h2o-llmstudio/train.py", line 749, in <module>
    run(cfg=cfg)
  File ./h2o-llmstudio/train.py", line 539, in run
    raise LLMTrainingException(
llm_studio.src.utils.exceptions.LLMTrainingException: System column 'i_dont_exist' not found in train DataFrame.
killed
```

closes #600

### DIFF
--- a/llm_studio/src/datasets/conversation_chain_handler.py
+++ b/llm_studio/src/datasets/conversation_chain_handler.py
@@ -147,8 +147,8 @@ class ConversationChainHandler:
         if cfg.dataset.system_column != "None":
             if cfg.dataset.system_column not in df.columns:
                 logger.warning(
-                    f"System column {cfg.dataset.system_column} not found."
-                    f"Disabling functionality."
+                    f"System column '{cfg.dataset.system_column}' not found."
+                    " Disabling functionality."
                 )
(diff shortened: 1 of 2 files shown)


psinger:
Not sure I am a big fan of adding single checks into the train.py, is this one really necessary? Sounds like a clear user error to me.

pascal-pfeiffer:
It adds quite some complexity to check for dataframe related issues. DF isn't loaded at the point where we check the cfg. So, we would need to change that. Plus, we need to pass the mode again (train, val). 
So, I decided to go for simple. If there are more checks, we can and should encapsulate it.

psinger (approved):

PR #856, opened 2024-09-13
Install flash-attention in the Docker image

ybabakhin:
Previously, flash-attention installation was failing with:
```
python3.10 -m pipenv run python -m pip install flash-attn==2.6.1 --no-build-isolation --upgrade --no-cache-dir
/workspace/.venv/bin/python3.10: No module named pipenv
```
e.g. https://github.com/h2oai/h2o-llmstudio/actions/runs/10838438509/job/30076644133

It was happening due to mixing the system and venv's Python interpreters. The fix is to add `pipenv` as a dependency to the Pipfile.
Successful build: https://github.com/h2oai/h2o-llmstudio/actions/runs/10848089540/job/30104506672


pascal-pfeiffer (commented):
Let's investigate if it isn't already enough to move this to after the setup

```Dockerfile
ENV PATH=/workspace/.venv/bin:$PATH
```

Issue #783, opened 2024-07-25
[DOC] Document Importing datasets with H2O Drive


sherenem:
### 📃 Documentation

Add feature documentation for importing datasets with H2O Drive

### Motivation

https://github.com/h2oai/h2o-llmstudio/pull/771

PR #859, opened 2024-09-18
Add importing dataset with h2o drive

oshi98:
This PR solves the Issue - https://github.com/h2oai/h2o-llmstudio/issues/783


pascal-pfeiffer (approved):
thank you!

PR #864, opened 2024-09-20
Pass settings for hf_transfer & chore

### LINKED ISSUE #801: [BUG] HF_Transfer is only applied to download of models
### 🐛 Bug

Follow up to #790 , HF_transfer should also be applied to upload of models when being set in the settings page. 

Currently, this is only the case when setting `HF_HUB_ENABLE_HF_TRANSFER="1"` as an ENV var.

### LLM Studio version
1.10.0-dev

pascal-pfeiffer:
- pass settings for hf_transfer, closes #801
- remove -C from train_wave.py and mark as deprecated in train.py
- HF_HUB_DISABLE_TELEMETRY=1 (afaik telemetry is only send in transformers examples, but won't hurt to have it there)
- New pipfile.lock for minor updates
- limit `markdown_cells` to object columns
- some renaming of methods for code clarity
- removed password option for settings in experiment config. We never show passwords/secrets on that page!
- renamed HUGGINGFACE_TOKEN to [default](https://huggingface.co/docs/huggingface_hub/package_reference/environment_variables#deprecated-environment-variables) HF_TOKEN
- some typing hints


ybabakhin (approved):
LGTM, thanks!

Issue #855, opened 2024-09-13
[UX] Missing ID column when a parent_id was specified, give a nice warning instead of a traceback


pascal-pfeiffer:
### 🔧 Proposed code refactoring

Instead of the rather cryptic error below, we should give a nice explanation what is missing in the dataset and how it can be fixed.

![image](https://github.com/user-attachments/assets/4e871631-9afa-4486-a044-e2da5cf0fd51)

### Motivation

UX

Chubop:
@us8945 Is this issue still open for contribution?

PR #865, opened 2024-09-20
Improve handling of dataset validation for Causal Language Models

us8945:
Resolve issue #855 
Changes:
- Add handling of Assert exception and improve user displayed message
- Improve assert errors to add information that will be relied back to user on how to fix the error

Test:
- When populating parent-id, select column that is populated for all records
- Try to Import dataset with no "id" column

### DIFF
--- a/llm_studio/app_utils/sections/dataset.py
+++ b/llm_studio/app_utils/sections/dataset.py
@@ -716,17 +716,6 @@ async def dataset_import(
             items = [ui.markup(content=header), ui.message_bar(text=text), plot_item]
             valid_visualization = True
 
-            # await busy_dialog(
-            #     q=q,
-            #     title="Performing sanity checks on the data",
-            #     text="Please be patient...",
-            # )
-            # add one-second delay for datasets where sanity check is instant
-            # to avoid flickering dialog
-            # Move sanity check to the beginning of step 5
-            # time.sleep(1)
-            # sanity_check(cfg)
-
         except AssertionError as exception:
             logger.error(f"Error while validating data: {exception}", exc_info=True)
             # Wrap the exception text to limit the line length to 100 characters


pascal-pfeiffer (dismissed):
Thank you @us8945 , I like the improved visuals.

Though, I think the process is still too difficult to understand for first time users. And the error message might now even be misleading. 

> Please go back and verify whether the problem type and other settings were set properly.

This pops up when I am setting up `parent_id` but let's say my `id` column is called `msg_id`. It isn't really clear for users how this can be fixed. I'd really go into detail explaining how the dataframe needs to be structured and set-up to be used in conversation chains. 

It also doesn't have to be on second page for such expected issues, but can be just an overlay that can be clicked away.

Might as well give the option to select another column as `id` column instead if it only is a naming issue. In that case, modify the dataframe before saving the dataset. And it will prevent that the user needs to go back to modify the dataframe outside of LLM Studio.

documentation/docs/guide/datasets/data-connectors-format.md
```diff
@@ -24,6 +24,7 @@ H2O LLM Studio supports the following data connectors to access or upload extern
     - **Prompt Column:** The column in the dataset containing the user prompt.
     - **Answer Column:** The column in the dataset containing the expected output.
     - **Parent Id Column:** An optional column specifying the parent id to be used for chained conversations. The value of this column needs to match an additional column with the name `id`. If provided, the prompt will be concatenated after preceding parent rows.
+    - **Id Column:**  The column is required only when the `Parent Id` is set. It identifies the record ID used for linked conversations. The value in the `Parent Id` column must match this column's value.
```
sherenem, line 27:
```suggestion
    - **Id Column:**  This column is only required when the `Parent Id` is set. It identifies the record ID used for linked conversations. The value in the `Parent Id` column must match this column's value.
```

documentation/docs/tooltips/experiments/_id-column.mdx
```diff
@@ -0,0 +1 @@
+The column is required only when the `Parent Id` is set. It identifies the record ID used for linked conversations. The value in the `Parent Id` column must match this column's value.
```
sherenem, line 1:
```suggestion
This column is only required when the `Parent Id` is set. It identifies the record ID used for linked conversations. The value in the `Parent Id` column must match this column's value.
```

sherenem (commented):
LGTM

documentation/docs/tooltips/experiments/_id-column.mdx
pascal-pfeiffer, line 1: Also needs to be added here to be listed in the documentation: https://github.com/h2oai/h2o-llmstudio/blob/main/documentation/docs/guide/experiments/experiment-settings.md
cc @sherenem

pascal-pfeiffer (changes requested):
Thank you for working on this issue. 

In the UI, changing the `parent_id` resets the state for other fields. This should not happen. See video below: 

https://github.com/user-attachments/assets/73eb7cc1-997f-495f-95ba-eb7f9fca60c6

If I change the ID column in the experiment settings now (while having a parent id set), the error is again too cryptic to understand:
```
ValueError: Found input variables with inconsistent numbers of samples: [13026, 12860]
```

Nitpick: the error is in a code field which is color coded and has a copy button. Neither is needed and adds distraction.
image

documentation/docs/tooltips/experiments/_id-column.mdx
pascal-pfeiffer: The issue surfaces when I am choosing the wrong ID column in the experiments settings. Prior, this wasn't an issue, as ID column was fixed and the dataset was validated against that. Now, we can change ID column when setting up a new experiment. 

image

pascal-pfeiffer (dismissed):
I am still getting the same error. The latest changes don't seem to get triggered.

```
ValueError: Found input variables with inconsistent numbers of samples: [13026, 12860]
```

This is just oasst defaults with `Parent_ID = parent_id` and setting a wrong `ID_Column = instruction`

image

PR #875, opened 2024-10-10
Change gcloud auth to OIDC

lakinduakash:
Previous gcp authenticate method ( with github action library ) save the key into github workspace. Since Dockerfile copy the current directory it copies the key as well. Enabling OIDC removes the dependency of the key file and uses short lives sessions and no key is stored.


hasithjp (approved):
LGTM thanks
Can you please add a brief PR description about this change

hasithjp (approved):

pascal-pfeiffer (approved):
lgtm

PR #891, opened 2025-01-16
[DOCS] Add pdf-docs workflow

sherenem:
This PR adds a GitHub workflow that generates PDF documentation of the LLM Studio docs so that our airgapped customers have a way to access offline documentation.

The workflow calls these two python scripts on the Makersaurus repo that process the doc .md files we currently have into a proper format that can be converted smoothly into a PDF:

- https://github.com/h2oai/makersaurus/blob/main/documentation/generate_combined_md.py
- https://github.com/h2oai/makersaurus/blob/main/documentation/generate_pdf_doc.py


VERSION
```diff
@@ -0,0 +1 @@
+1.13.0
```
pascal-pfeiffer, line 1: can we fetch from https://github.com/h2oai/h2o-llmstudio/blob/main/pyproject.toml#L3 instead to have a single source of truth?

pascal-pfeiffer: Or alternatively, move to /documentation/VERSION, and it will be your responsibility to update for new documentation versions.

sherenem: @pascal-pfeiffer I think the issue would be that - not all doc repos have that `pyproject.toml` file and since we are pulling the same script from a single source for docs across all other products as well, I am not sure if we could change this separately for LLM Studio.

pascal-pfeiffer: Then consider the alternative above, please

VijithaEkanayake: @sherenem @pascal-pfeiffer Version is an argument that we pass into the common script, We can fetch the version from https://github.com/h2oai/h2o-llmstudio/blob/main/pyproject.toml#L3 and pass it to the script

sherenem: @pascal-pfeiffer this has been changed now with commit https://github.com/h2oai/h2o-llmstudio/pull/891/commits/23c01287fd13fe2da62d752f8049996d3ac2a530

pascal-pfeiffer (approved):
Lgtm

PR #903, opened 2025-03-03
Bump Makersaurus version

sherenem:
This PR upgrades the doc platform (Makersaurus) to v0.11.1 and fixes any formatting conflicts.

### DIFF
--- a/documentation/docs/faqs.md
+++ b/documentation/docs/faqs.md
@@ -1,7 +1,7 @@
 ---
 description: Learn about frequently asked questions. 
 ---
-import Icon from "@material-ui/core/Icon";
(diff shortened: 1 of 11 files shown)


5675sp (approved):
I pushed a few changes because a missing tooltip was preventing the successful development of the documentation locally. After those changes, I built and served the documentation successfully. Please review my changes to ensure I added the correct tooltip on the "Experiment settings" page.  

Side note: When running "make setup-doc," we get a few notices that specific packages need to be updated. Despite that warning, the documentation is still being built successfully. It might be fine for now, but it might break in the future. I already notified Ben about the notifications. We will explore.

PR #905, opened 2025-04-03
Update dependencies, cuda

pascal-pfeiffer:
- update cuda to 12.4
- update pip/pipenv and python packages
- replace sqlitedict with diskcache
- resolves CVEs
- fix mypy checks


dott1718 (approved):
lgtm

PR #927, opened 2025-08-01
Add internal ECR build infrastructure

codyharris-h2o-ai:
Since we do not have the ability to pull from the managed cloud ECR, here is a github action change to publish images to the internal ECR as well.


pascal-pfeiffer (approved):
lgtm

PR #936, opened 2025-09-18
Add missing ecr login for internal bundle publish and refactor build steps

codyharris-h2o-ai:
* Only run build on tags with `v` prefix
* Add missing ecr login step for internal bundle publish
* Refactor to perform bundle step once and share the artifact between the jobs
* For internal publish, only check out the required Dockerfile instead of the entire repository


pascal-pfeiffer (approved):
thank you! lgtm

PR #1062, opened 2026-06-03
ci: use OIDC role secret for OIDC role in workflows

dhanikaa:
## Summary

Replaces the hardcoded AWS OIDC role ARN in the test workflows with the `GH_OIDC_ROLE_LLM_STUDIO_ENTERPRISE` repository secret.

Previously the role was hardcoded in four places. Moving it to a secret avoids leaking the account ID/role in source and lets the role be rotated without code changes.

## Changes

- `.github/workflows/test.yml` — `role-to-assume` in the start-runner and stop-runner jobs now reference `${{ secrets.GH_OIDC_ROLE_LLM_STUDIO_ENTERPRISE }}`
- `.github/workflows/test_ui.yml` — same change in both jobs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,7 @@ jobs:
         with:
           aws-region: us-east-1
           role-session-name: h2o-llmstudio-enterprise-ec2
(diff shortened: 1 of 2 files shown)


pascal-pfeiffer (approved):
lgtm, but please fix the creds

PR #1099, opened 2026-08-13
fix: don't let an optional training dependency break app startup

avnerv:
## Problem

LLM Studio shows a blank screen on some deployments — the app dies during import, so nothing ever listens on the port:

```
File ".../deepspeed/ops/transformer/inference/triton/matmul_ext.py", line 37, in is_nfs_path
  fs_type = lines[1].split()[1].lower()  # File system type is the second column
IndexError: list index out of range
```

## Root cause

`is_nfs_path()` runs `df -T <path>` and reads the filesystem type from the second column of the second line:

```python
output = subprocess.check_output(['df', '-T', path], encoding='utf-8')
lines = output.strip().split('\n')
if len(lines) > 1:                          # lines[1] IS guarded
    fs_type = lines[1].split()[1].lower()   # .split()[1] is NOT
```

BusyBox `df` wraps a device name longer than ~20 characters onto its own line, so `lines[1]` holds only the device, `.split()` returns one element, and `[1]` raises. The image ships BusyBox, not coreutils. This runs at **module import scope**, so the whole app dies — and the call only exists to print an informational NFS warning.

`TRITON_CACHE_DIR` points at `/mount`, an EBS-backed volume. Some volumes are presented as `/dev/nvme6n1` (12 chars, fine) and others as `/dev/dvol05d356c8837b6754f` (25 chars — `d` + the EBS volume id, which wraps). Observed **both on the same node and instance type**, so it is a property of how the individual volume was attached, not of the node. That is why this looks random and is hard to reproduce on demand.

It is also why a fresh instance usually boots fine: on a new volume the cache directory doesn't exist yet, `df` on a missing path exits non-zero, and `CalledProcessError` is caught and returns harmlessly. DeepSpeed then creates the directory, so only a **later restart** crashes — which is how affected releases pass QA.

Upstream issue: https://github.com/deepspeedai/DeepSpeed/issues/8249

## This PR

**`patches/deepspeed-0.17.5-df-parse.patch`** — `df -T` → `df -PT`. POSIX `-P` guarantees one line per filesystem and suppresses the wrap. Applied at image build time (`Dockerfile`, right after `uv sync`), with a `Makefile` target for local dev, following the `patches/` layout used in h2oai/htab. Removed once the fix is released upstream.

**`modeling_utils.py`** — move the `deepspeed` imports into the `use_deepspeed` branches that already guard every use of them. That file is in the app's UI import chain, so an optional *training* dependency could take the whole app down at startup. `train.py` keeps its module-scope import.

## Validation

On a node exhibiting the long device name, with `TRITON_CACHE_DIR=/mount/.triton/cache`:

- **unpatched:** `import deepspeed` → `IndexError: list index out of range`
- **with `df -PT`:** import proceeds past `is_nfs_path`
- **`df -T` → second line has 1 field; `df -PT` → 7 fields**

The patch was also verified to apply cleanly against real DeepSpeed 0.17.5 source and to be idempotent on re-application.

Note: getting a clean `import deepspeed` also required an environment fix on our side, unrelated to this PR — the runtime ran as a uid with no `/etc/passwd` entry, so torch's inductor `cache_dir()` hit `KeyError: getpwuid()` via `getpass.getuser()`. That is fixed in our deployment, not here, but it means the DeepSpeed patch alone was not sufficient to bring the app up in our environment.

## Affected versions

All releases from **v1.14.4** (where the cache moved to `/mount` in #935) through **v1.14.15**. v1.14.3 and earlier are unaffected. Rolling back isn't viable — FedRAMP needs 1.14.15 for CVEs. Upgrading DeepSpeed doesn't help; the parse is unchanged through 0.19.5.


pascal-pfeiffer (approved):
thank you, lgtm!

